PR #4, opened 2026-05-01
Feat/extra route paths

amrashrafh1:
## Problem                                                                                                                                          
                                                                                                                                                      
  `RouteAnalyzer::findRouteFiles()` hard-codes the top-level `routes/` directory as the only scan root. Routes registered from service providers      
  outside that directory — common in modular, DDD, and package-style Laravel projects — are silently ignored.                                         
                                                                                                             
  For example, none of these are discovered today:                                                                                                    
  - `app/Modules/Auth/routes/api.php`                                                                                                                 
  - `app/Domain/Billing/routes/api.php`
                                                                                                                                                      
  ## Solution                                                                                                                                         
             
  Add an optional published config (`config/laravel-brain.php`) with a `route_paths` key that accepts glob patterns relative to the project root. The 
  standard `routes/` directory is always scanned; extra patterns are additive and deduplicated.                                                       
                                                                                               
  ```php                                                                                                                                              
  // config/laravel-brain.php                            
  return [                   
      'route_paths' => [
          'app/Modules/*/routes/*.php',
      ],                               
  ];                                                                                                                                                  
    
  No breaking changes — projects that don't publish the config get identical behaviour.                                                               
                                                                                                                                                      
  Changes
                                                                                                                                                      
  - config/laravel-brain.php — new config file with route_paths key                                                                                   
  - LaravelBrainServiceProvider — merges config defaults, publishes via laravel-brain-config tag
  - RouteAnalyzer::analyze() — accepts $extraGlobs parameter                                                                                          
  - RouteAnalyzer::findRouteFiles() — scans extra globs with deduplication                                                                            
  - ProjectAnalyzer::analyze() — reads config('laravel-brain.route_paths') and threads it through                                                     
  - tests/fixtures/.../app/Modules/Auth/routes/api.php — new fixture for module-style routes                                                          
  - RouteAnalyzerTest — 3 new tests: no extra globs (regression), module discovery, overlap dedup                                                     
  - README.md — updated route discovery section with usage example

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,6 +114,24 @@ GET /_laravel-brain
 
 LaravelBrain recursively scans your entire `routes/` directory — not just `web.php` and `api.php`. Any PHP file under `routes/**` is analyzed, including versioned files like `routes/v1/users.php` or module-specific files like `routes/modules/admin.php`.
 
+For **modular, DDD, or package-style projects** where routes are registered from service providers outside the `routes/` directory, publish the config and add glob patterns:
+
+```bash
+php artisan vendor:publish --tag=laravel-brain-config
+```
+
+```php
+// config/laravel-brain.php
+return [
+    'route_paths' => [
+        'app/Modules/*/routes/*.php',
+        // 'app/Domain/*/routes/*.php',
+    ],
+];
+```
+
+Patterns are relative to the project root and merged with the standard `routes/` scan. Overlapping paths are deduplicated automatically.
+
 ### Call chain tracing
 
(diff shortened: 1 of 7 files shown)


jdecode:
@amrashrafh1 
Would it be better/possible to get the list of routes using the command `php artisan route:list --except-vendor` instead?
This would mean that irrespective of the path of the route files, those routes can still be accessed?
Or perhaps there would be an underlying Laravel method to expose all details of all routes (the method that is used by the route:list command).

PR #29, opened 2026-05-08
Remove hard limitation when using local memory init

adepane:
When I tried to scan with big project, unfortunately this line as blocker even though I tried to use from `php -d memory_limit=-1`, rather than put hard limit, better use limit on artisan or set from the php.ini

### DIFF
--- a/src/Commands/ScanCommand.php
+++ b/src/Commands/ScanCommand.php
@@ -23,7 +23,7 @@ class ScanCommand extends Command
     {
         $memoryLimit = $this->normalizeMemoryLimit($this->option('memory-limit'));
 
-        if($memoryLimit === self::FAILURE) {
+        if ($memoryLimit === self::FAILURE) {
             return $memoryLimit;
         }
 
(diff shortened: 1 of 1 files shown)


mrmarchone:
Nice but do i think it would be better if we set it as a command argument ? @adepane

adepane:
@mrmarchone yes sure it is good idea, I already implemented. By default this argument will not receive least than 1024M, so 1024M as default as like yours idea before.

This docs I write related the changes.

### Memory Management in Laravel Brain
The `brain:scan` command includes a `--memory` option to control the PHP memory limit during the analysis process. This is particularly useful for large projects where the default PHP limit might be insufficient.

### Usage
```
php artisan brain:scan --memory=2048
```

### Configuration Details
**Default Value**: `1024M`
**Base Unit**: Megabytes (`M`). If you provide a raw number (e.g., `512`), it is automatically treated as `512M`.

### Supported Format:
**Numeric (MB)**: `512`, `2048` (interpreted as `512M`, `2048M`)
**With Units**: `2G`, `4G`, `1024M`
**Unlimited**: `-1`
**Validation**: The command will validate the input. If a non-positive number (other than `-1`) is provided, the command will display an error and stop execution.

mrmarchone:
Great and appreciated your work thanks @adepane

mrmarchone:
Great work, could you please open this PR for main branch ?

src/Commands/ScanCommand.php
```diff
@@ -11,7 +11,8 @@ class ScanCommand extends Command
 {
     protected $signature = 'brain:scan
                             {--watch : Watch for PHP file changes and auto-rescan}
-                            {--interval=3 : Poll interval in seconds (watch mode only)}';
+                            {--interval=3 : Poll interval in seconds (watch mode only)}
+                            {--memory=1024M : Increase memory limit for scanning. Example: 1024M}';
```
mrmarchone: name it --memory-limit

src/Commands/ScanCommand.php
```diff
@@ -20,7 +21,13 @@ class ScanCommand extends Command
 
     public function handle(): int
     {
-        ini_set('memory_limit', '1024M');
+        $memory = $this->setMemoryLimit($this->option('memory'));
```
mrmarchone: This method does not “set” the memory limit. call it $memory = $this->normalizeMemoryLimit($this->option('memory'));

src/Commands/ScanCommand.php
```diff
@@ -116,6 +123,55 @@ private function summariseChanged(array $changed): string
...
+
+    private function setMemoryLimit($option): string|int
+    {
+        $memory = trim((string) $option);
+
+        if ($memory === '-1') {
+            return -1;
+        }
+
+        if (preg_match('/^\s*([+-]?\d+)\s*([kmgt]?)\s*$/i', $memory, $matches)) {
```
mrmarchone: could you try this regex '/^([+-]?\d+)([kmgt]?)$/i'

src/Commands/ScanCommand.php
mrmarchone, line 1: Review this one please , this an enahced version
```
private function normalizeMemoryLimit($option): string|int
{
    $memory = strtoupper(trim((string) $option));

    if ($memory === '-1') {
        return -1;
    }

    if (! preg_match('/^(\d+)([KMGT]?)$/', $memory, $matches)) {
        $this->error('Invalid memory limit format. Example: 1024, 1G, 2G or -1.');

        return self::FAILURE;
    }

    $value = (int) $matches[1];
    $unit = $matches[2] ?: 'M';

    if ($value <= 0) {
        $this->error('Invalid memory limit. The value must be a positive number or -1.');

        return self::FAILURE;
    }

    if ($this->convertToBytes($value, $unit) < 1024 ** 3) {
        $this->error('The memory limit must be at least 1024M (or equivalent).');

        return self::FAILURE;
    }

    return "{$value}{$unit}";
}
```

src/Commands/ScanCommand.php
```diff
@@ -116,6 +123,55 @@ private function summariseChanged(array $changed): string
...
+    private function convertToBytes(int $value, string $unit): int
+    {
+        return match ($unit) {
+            'K' => $value * 1024,
+            'M' => $value * 1024 * 1024,
+            'G' => $value * 1024 * 1024 * 1024,
+            'T' => $value * 1024 * 1024 * 1024 * 1024,
+            default => $value,
+        };
+    }
```
mrmarchone, line 169: Could you try this also ?
```
private function convertToBytes(int $value, string $unit): int
{
    return match ($unit) {
        'K' => $value * (1024 ** 1),
        'M' => $value * (1024 ** 2),
        'G' => $value * (1024 ** 3),
        'T' => $value * (1024 ** 4),
        default => $value,
    };
}
```

mrmarchone (changes requested):

adepane:
> Great work, could you please open this PR for main branch ?

sure, I will try to update your request

mrmarchone:
Please solve the pipeline @adepane

adepane:
> Please solve the pipeline @adepane

done, thanks.

mrmarchone:
> > Please solve the pipeline @adepane
> 
> done, thanks.

It merged on main branch , i will release a new version soon , keep your eye 💪

PR #41, opened 2026-05-15
Add Laravel Boost skill

SanderMuller:
### What

Adds a Laravel Boost skill, following the same pattern as `livewire/blaze` (`resources/boost/skills/blaze-optimize/`). The skill is agent-agnostic — `boost:update` distributes it to whichever AI tools the developer has configured (Claude Code, Cursor, GitHub Copilot, etc.).

```
resources/boost/skills/laravel-brain/
├── SKILL.md                        # Step-by-step analysis workflow
└── references/
    ├── security-patterns.md        # Anti-pattern catalogue with fixes
    └── interpreting-output.md      # Guide to reading brain:export-context output
```

### Why

`brain:generate-rules` generates static context files for specific agents (Cursor, Windsurf, JetBrains Junie, OpenAI Codex). A Boost skill is different — it provides structured workflow guidance that tells the agent *how* to use the tool, not just that it exists.

Without a skill, agents default to `brain:export-context` with the 6000-token budget, which silently truncates the hotspot table on any project with more than ~50 controllers. They also skip the security category breakdown, since that data is only available via the viewer's JSON API (`/_laravel-brain/api/context?format=json`), not in the CLI export.

### What it does

- Installs the package if missing, then runs `brain:scan` and calibrates the export budget to graph size
- Fetches per-category security counts from the viewer JSON API before reading any source
- Filters Filament/query-builder false positives from the hotspot table; ranks by `cyclomatic × lines`
- Reads the top candidates in source and checks for 8 specific anti-patterns
- Runs `brain:export-context --node` on confirmed findings to verify middleware before writing fixes
- Outputs a structured report (Security / Complexity / DB concerns) with only source-verified items

### Design decisions worth noting

**Budget calibration.** The skill always overrides the default budget based on graph size (8k / 15k / 20k tokens for <300 / 300–800 / 800+ nodes). Discovered when the default truncated the entire hotspot table on a real project.

**JSON API security breakdown.** The `brain:scan` surface count is a single number. The skill fetches per-category counts from the viewer API before reading any source — this was the highest-signal step in practice, revealing 125 `PUBLIC_WRITE` findings the plain export never surfaced.

**CLI limitation documentation.** The skill calls out that `--node` / `--route` filters only affect the Call Chain section, not the Complexity Hotspots table. This is non-obvious and caused confusion when trying to scope exports.

### Open question

Should `brain:generate-rules` also output the Boost skill alongside its existing per-agent outputs? Happy to add that in this PR or leave it as a follow-up — the skill works standalone either way.

### DIFF
--- a/resources/boost/skills/laravel-brain/SKILL.md
+++ b/resources/boost/skills/laravel-brain/SKILL.md
@@ -1,0 +1,257 @@
+---
+name: laravel-brain
+description: "Analyze a Laravel application for complexity hotspots, security surface, and database access patterns using laravel-brain. Activate when: the user asks what needs work, what should be improved, what to work on next, where to focus, what is the most complex code, what are the security issues, audit the codebase, analyze code quality, find technical debt, find refactoring targets, what are the worst parts of the codebase, or when user mentions: brain, hotspots, cyclomatic, complexity audit, security audit, codebase health, codebase review."
+---
+
+# Laravel Brain Analysis
+
+`laravel-brain` scans the application graph and surfaces cyclomatic complexity hotspots, security surface area, and database access patterns across every controller, command, and Livewire component. Use it to answer "what needs work" before writing code, not after.
+
+EXECUTE steps 0–5 IMMEDIATELY without asking the user. Scans are read-only and safe. Only pause at step 6 to present findings.
+
+---
+
+## 0. Pre-flight
+
+Ensure the package is installed before scanning:
+
+```bash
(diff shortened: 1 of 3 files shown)


mrmarchone:
Thanks @SanderMuller , i will review this PR ASAP.
Could you please change it on main branch ?

SanderMuller:
> Thanks @SanderMuller , i will review this PR ASAP. Could you please change it on main branch ?

Done!

PR #46, opened 2026-05-17
Refactor RouteAnalyzer to enhance path resolution

mrmarchone:
### Problem

When routes are registered using:

```php
Route::prefix('api/admin')
    ->group(base_path('routes/admin.php'));
```

the package cannot reliably determine the original route file (`admin.php`, `restaurant.php`, etc.) because Laravel stores the controller action instead of the route file path in the route action metadata.

This causes incorrect or missing route grouping/filtering when trying to identify routes by file source.

### Solution

This PR improves route detection by using the generated route URI/prefix instead of relying on the `uses` action metadata.

Example:

* `api/admin/*` → `routes/admin.php`
* `api/restaurant/*` → `routes/restaurant.php`

### Benefits

* Works correctly with Laravel 13 route registration style
* Supports grouped route files with prefixes
* Fixes route filtering for modular route structures

### Tested

Tested with:

* Laravel 13
* Multiple grouped route files
* Prefixed API routes


mrmarchone:
Thanks @hussein4alaa , could you please fix the pipeline ?

mrmarchone:
We need also the unit tests for it, could you please adding the tests ?

PR #50, opened 2026-06-18
feat(security): reduce false positives in SecurityAnalyzer

tecnomanu:
On a real Laravel 13 app (131 routes, 46 controllers) `php artisan brain:scan` currently reports 18 security issues — all of them false positives — by way of three independent rules in `SecurityAnalyzer.php`. The noise buries the handful of genuine findings and makes the security tab hard to trust. This PR fixes the three root causes with focused changes and 12 new tests.

### Bug 1 — `UNVALIDATED_INPUT` matched any `->all()`

Pass 2 of `scanAstNode()` walked every `MethodCall` named `all()` regardless of receiver, so each of these legitimate calls became a high-severity finding:

\`\`\`php
collect((array) data_get($t->settings, 'methods', []))->values()->all()
$query->get(['col'])->all()
collect($e->getTrace())->take(8)->all()
\`\`\`

The visitor now only fires when the receiver is provably an HTTP Request — either a method-parameter typed as \`Illuminate\\Http\\Request\` (or any FormRequest-shaped class) or the bare \`request()\` helper. The check is promoted to a public static \`SecurityAnalyzer::isRequestAllCall()\` so the existing Pass-1 mass-assignment visitor reuses the same receiver logic.

### Bug 2 — `signed`, custom HMAC, and webhook routes flagged as `PUBLIC_WRITE`

Webhooks (`POST /webhooks/stripe/{tenant?}`), token-in-URL magic-link endpoints (`POST /portal/{token}/cancel`), and custom-merchant-API routes all enforce authentication outside the standard `auth:*` middleware. With no extensibility, the only fix was to ignore the entire SecurityAnalyzer.

This PR:

- Adds `signed` and `Illuminate\Routing\Middleware\ValidateSignature` to `AUTH_PATTERNS` — a signed URL is cryptographic authentication of the request.
- Adds a `security` block to `config/laravel-brain.php` so apps can declare their own auth-providing middleware and trust webhook URI globs:

  \`\`\`php
  'security' => [
      'auth_middleware_patterns'     => ['merchant.hmac'],
      'throttle_middleware_patterns' => [],
      'trusted_route_names'          => ['webhooks.*'],
      'trusted_route_uris'           => ['webhooks/*', 'portal/*'],
  ],
  \`\`\`

- Defaults are empty, so no behaviour change for apps that don't publish the config — except for the `signed`-as-auth fix, which is a strict improvement.

### Bug 3 — Breeze `POST /login` etc. flagged as `MISSING_THROTTLE`

Breeze 2.x / Fortify enforce login throttling via `RateLimiter::tooManyAttempts(...)` inside the controller action (or inside the bound `LoginRequest::ensureIsNotRateLimited()`) — not via the `throttle:` middleware. The current middleware-only check produces a high-severity finding on every Breeze auth route.

The analyzer now also scans:

- The controller method AST for `RateLimiter::tooManyAttempts/hit/for/attempt/availableIn`, `$x->tooManyAttempts()`, or `$x->hitRateLimit()`.
- Each `FormRequest` typed in the action's parameters (so `LoginRequest::ensureIsNotRateLimited()` is reached).
- The controller class for the `Illuminate\Foundation\Auth\ThrottlesLogins` trait.

When any of these is found, `MISSING_THROTTLE` is suppressed.

## Real-world impact

A Laravel 13 SaaS app, 131 routes, 46 controllers:

| Setup | UNVALIDATED_INPUT | PUBLIC_WRITE | Total |
| --- | --- | --- | --- |
| `v2.3.0` | 5 | 13 | 18 |
| After this PR, no config | 0 | 13 | 13 |
| After this PR + `trusted_route_uris` published | 0 | 0 | 0 |

All eliminated findings were verified by hand as false positives; no real issue was hidden.

## Tests

12 new Pest tests in `tests/Unit/SecurityAnalyzerTest.php` cover each bug both ways — the FP suppression AND the still-flagged regressions:

\`\`\`
PASS  Tests\\Unit\\SecurityAnalyzerTest
  ✓ does not flag Collection->values()->all() as UNVALIDATED_INPUT
  ✓ does not flag Eloquent Collection ->all() as UNVALIDATED_INPUT
  ✓ still flags request()->all() (helper) as UNVALIDATED_INPUT
  ✓ treats `signed` middleware as authentication, not public
  ✓ treats `ValidateSignature` FQCN middleware as authentication
  ✓ treats configured auth_middleware_patterns as authentication
  ✓ suppresses PUBLIC_WRITE on routes matching trusted_route_names glob
  ✓ suppresses PUBLIC_WRITE on routes matching trusted_route_uris glob
  ✓ still flags public mutating routes that are NOT trusted
  ✓ does not flag MISSING_THROTTLE when controller uses RateLimiter::tooManyAttempts
  ✓ does not flag MISSING_THROTTLE when the action FormRequest throttles
  ✓ still flags MISSING_THROTTLE on sensitive routes with no rate-limiting
\`\`\`

Full suite: 82 passed (was 70), 0 phpstan errors, pint clean.

## Compatibility

- No public API breaks — `SecurityAnalyzer::__construct` takes an optional `?array $config` for tests but defaults to reading from `config('laravel-brain.security')` exactly as the rest of the package does.
- Adding `signed` to `AUTH_PATTERNS` is the only default-behaviour change. It is a fix, not a regression.
- Config defaults are empty arrays, so apps that don't publish `laravel-brain.php` see identical behaviour for the rest of the rules.
- No new runtime dependencies; nothing changes for the Laravel 9 / 10 / 11 / 12 / 13 matrix.

## Files

- `src/Analysis/SecurityAnalyzer.php:39` — added `signed` + `ValidateSignature` to AUTH_PATTERNS; new constructor + config plumbing; `isTrustedRoute()`, `hasInCodeThrottle()`, `astHasRateLimiterCall()`, `astUsesThrottlesLoginsTrait()`, `resolveFormRequestFiles()`; static `isRequestAllCall()`; FunctionLike-aware `collectRequestParamNames()`.
- `config/laravel-brain.php` — new `security` block, all defaults empty.
- `tests/Unit/SecurityAnalyzerTest.php` + `tests/fixtures/security-project/` — 12 tests + 5 fixture files.

### DIFF
--- a/config/laravel-brain.php
+++ b/config/laravel-brain.php
@@ -90,20 +90,63 @@ return [
     // -------------------------------------------------------------------------
     // Security Analysis
     // -------------------------------------------------------------------------
-    // Extra middleware aliases or FQCN prefixes that Brain should treat as
-    // authentication middleware. Entries are matched as case-insensitive
-    // prefixes, so 'auth.custom' matches 'auth.custom:api'.
+    // Override / extend the SecurityAnalyzer heuristics that classify each
+    // route's exposure and decide what counts as throttled or trusted.
     //
-    // Add your application's custom auth middleware here to prevent
-    // false "Public Route" warnings.
+    // The analyzer always starts from a conservative built-in default list
+    // and *adds* anything declared here on top — it never replaces the
+    // defaults. Entries are matched as case-insensitive prefixes; a plain
+    // name like 'auth.custom' matches 'auth.custom' and 'auth.custom:api'.
     //
-    // Example:
-    //   'auth_middleware' => [
-    //       'App\\Http\\Middleware\\CustomAuth',
-    //   ],
+    // When in doubt, leave everything empty. The defaults catch the
+    // standard Laravel middleware (`auth`, `auth:sanctum`, `signed`,
+    // `throttle`, etc.).
     //
     'security' => [
+
+        // Extra middleware aliases or FQCN prefixes that Brain should treat
+        // as authentication. Add your application's custom HMAC / api-key
+        // / bearer-token middleware here so routes guarded by it stop
+        // being reported as "public" / PUBLIC_WRITE.
+        //
+        // Examples:
+        //   'auth.custom',
+        //   'merchant.hmac',
+        //   App\Http\Middleware\AuthenticateMerchant::class,
+        //
         'auth_middleware' => [],
+
+        // Extra middleware aliases or FQCN prefixes that Brain should treat
+        // as rate-limiting. The default list already covers `throttle:` and
+        // ThrottleRequests; add any custom throttle middleware here.
+        //
+        'throttle_middleware' => [],
+
+        // Route names whose routes are trusted (PUBLIC_WRITE will be
+        // suppressed even when no recognised auth middleware is present).
+        //
+        // Use this for endpoints whose authentication is enforced *outside*
+        // of Laravel's middleware system — typically a webhook with HMAC
+        // verification done in the controller, or an endpoint authenticated
+        // by a token in the URL.
+        //
+        // Supports glob patterns (`*` and `?`), matched case-insensitively.
+        //
+        // Examples:
+        //   'webhooks.*',
+        //   'portal.*',
+        //
+        'trusted_route_names' => [],
+
+        // Route URI globs whose routes are trusted, same semantics as
+        // `trusted_route_names`. Matched against the route's URI without
+        // a leading slash. Useful when routes are unnamed.
+        //
+        // Examples:
+        //   'webhooks/*',
+        //   'portal/*/cancel',
+        //
+        'trusted_route_uris' => [],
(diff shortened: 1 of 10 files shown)


mrmarchone:
Thanks for your support, could you please change the pr to main branch?

tecnomanu:
Absolutely — already done! Rebased onto `main` and force-pushed. The PR now builds on the new `extraAuthPatterns` infra from #49 instead of introducing parallel config keys: I just added `extraThrottlePatterns`, `trustedRouteNames`, and `trustedRouteUris` as additional constructor params next to the existing one, extended the `security` config block accordingly, and appended my 11 tests to the `SecurityAnalyzerTest.php` you created (didn't override it).

Full suite: 87/87 passing, phpstan clean, pint clean. Let me know if you'd like any adjustments :)

PR #51, opened 2026-06-18
fix(security): recognise custom auth guards by their extends chain

tecnomanu:
Rebased onto current `main` and narrowed to the classifier change. The array-form `$middleware->alias([...])` fix is split out to #86 (per @SanderMuller's suggestion), and the review items that only applied to the old base — the `SAFE_INFRA_PATTERNS` suppression, the duplicated locate helper, the `signed`/`ValidateSignature` contradiction — are gone with the rebase (current `main` never had them).

## Problem

`classifyExposure()` already recognises the framework `Authenticate` and an app subclass that happens to **share its name** (`App\Http\Middleware\Authenticate`), via basename matching. It still misses a custom guard whose class name doesn't contain "auth" — the textbook second-guard case:

```php
$middleware->alias(['auth.customer' => AuthenticateCustomer::class]);
```

`AuthenticateCustomer extends Authenticate`, but its basename isn't `authenticate`, so the route is classified `public` and every mutating route behind it gets a false `PUBLIC_WRITE`.

## Change

`classifyExposure()` now also does a **verified `extends Authenticate` walk** (reusing the existing `PhpExtendsFqcnResolver`) for any middleware that pattern matching didn't already resolve. A guard that reaches `Illuminate\Auth\Middleware\Authenticate` through its parent chain is classified `authed`, which in turn suppresses the false `PUBLIC_WRITE`.

The walk is memoised per scan. Addressing your review: the memo is **seeded before recursing**, so a cyclic `extends` chain (a broken class that Brain can't execute) terminates instead of exhausting memory and taking the whole scan down.

No public API change; no new constructor parameters.

## Tests

Appended to `tests/Unit/SecurityAnalyzerTest.php` (+ `custom-auth-project` middleware fixtures):

- a differently-named `Authenticate` subclass resolved via an alias → `authed`, no `PUBLIC_WRITE`
- the same subclass given by bare FQCN → `authed`
- a cyclic `A extends B extends A` chain terminates (route stays `public`, `PUBLIC_WRITE` still fires)

Full suite green (213 passed), phpstan 0 errors, pint clean.

## Relationship to #86

Independent and self-contained (its tests build the alias map directly), but they compose: #86 makes the array-form alias reach the registry, and this PR classifies the resolved guard correctly. Either order merges cleanly.

### DIFF
--- a/src/Analysis/SecurityAnalyzer.php
+++ b/src/Analysis/SecurityAnalyzer.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace LaraMint\LaravelBrain\Analysis;
 
+use LaraMint\LaravelBrain\Parser\PhpExtendsFqcnResolver;
 use LaraMint\LaravelBrain\Parser\PhpFileParser;
 use PhpParser\Node;
 use PhpParser\NodeTraverser;
@@ -105,6 +106,14 @@ class SecurityAnalyzer
 
     /** @var array<string, list<array<string, mixed>>>|null */
     private ?array $externalByFile = null;
+
+    /**
+     * Memoised result of {@see self::isAuthClass()} keyed by FQCN, so each
+     * middleware class's `extends` chain is walked at most once per scan.
+     *
+     * @var array<string, bool>
+     */
+    private array $authClassCache = [];
 
     /**
      * Effective auth-middleware patterns (defaults + $extraAuthPatterns),
@@ -281,9 +290,16 @@ class SecurityAnalyzer
                 return 'admin';
             }
         }
-        // Auth check
+        // Auth check — pattern/basename match OR a verified `extends
+        // Authenticate` chain. The chain walk catches a custom guard whose
+        // class name doesn't contain "auth" (e.g. `auth.customer` mapped to
+        // App\Http\Middleware\AuthenticateCustomer), which basename matching
+        // against the framework `Authenticate` alone would miss.
         foreach ($middlewares as $mw) {
             if ($this->middlewareMatches($mw, $this->authPatterns)) {
+                return 'authed';
+            }
+            if ($this->isAuthClass($this->middlewareClass($mw))) {
                 return 'authed';
             }
         }
@@ -593,6 +609,94 @@ class SecurityAnalyzer
         $pos = strrpos($name, '\\');
 
         return strtolower($pos === false ? $name : substr($name, $pos + 1));
+    }
+
+    /**
+     * The class/alias part of a resolved middleware string, stripped of any
+     * `:params` suffix. The registry has already resolved aliases to FQCNs,
+     * so this is usually a class name.
+     */
+    private function middlewareClass(string $middleware): string
+    {
+        return explode(':', $middleware, 2)[0];
+    }
+
+    /**
+     * True when $fqcn is the framework's Authenticate middleware or any class
+     * whose `extends` chain reaches it. Results are memoised per scan so each
+     * FQCN is walked at most once.
+     */
+    private function isAuthClass(string $fqcn): bool
+    {
+        $fqcn = ltrim($fqcn, '\\');
+        if ($fqcn === '') {
+            return false;
+        }
+        if (isset($this->authClassCache[$fqcn])) {
+            return $this->authClassCache[$fqcn];
+        }
+        if ($fqcn === 'Illuminate\\Auth\\Middleware\\Authenticate') {
+            return $this->authClassCache[$fqcn] = true;
+        }
+
+        // Seed the memo *before* recursing so a cyclic `extends` chain (two
+        // classes that extend each other — only possible with a broken class
+        // Brain can't execute) terminates instead of exhausting memory and
+        // taking the whole scan down. The real result overwrites this below.
+        $this->authClassCache[$fqcn] = false;
+
+        $parent = $this->resolveParentClass($fqcn);
+        if ($parent === null) {
+            return false;
+        }
+
+        return $this->authClassCache[$fqcn] = $this->isAuthClass($parent);
+    }
+
+    /**
+     * Best-effort: parse the file declaring $fqcn (PSR-4 convention) and
+     * return the FQCN of its `extends` clause, or null when the file can't be
+     * found or the class doesn't extend anything we can resolve.
+     */
+    private function resolveParentClass(string $fqcn): ?string
+    {
+        $file = $this->locateClassFile($fqcn);
+        if ($file === null) {
+            return null;
+        }
+
+        $parsed = $this->cachedParse($file);
+        if (($parsed['ast'] ?? null) === null) {
+            return null;
+        }
+
+        $namespace = PhpExtendsFqcnResolver::namespaceFromAst($parsed['ast']);
+        $useMap = $parsed['useMap'] ?? [];
+
+        $parent = null;
+        $traverser = new NodeTraverser;
+        $traverser->addVisitor(new class($parent, $namespace, $useMap) extends NodeVisitorAbstract
+        {
+            public function __construct(public ?string &$parent, private string $namespace, private array $useMap) {}
+
+            public function enterNode(Node $node): ?int
+            {
+                if ($node instanceof Node\Stmt\Class_ && $node->extends !== null) {
+                    $this->parent = PhpExtendsFqcnResolver::resolveExtends(
+                        $node->extends,
+                        $this->namespace,
+                        $this->useMap,
+                    );
+
+                    return NodeTraverser::STOP_TRAVERSAL;
+                }
+
+                return null;
+            }
+        });
+        $traverser->traverse($parsed['ast']);
+
+        return $parent;
     }
 
     /**
--- a/tests/Unit/SecurityAnalyzerTest.php
+++ b/tests/Unit/SecurityAnalyzerTest.php
@@ -487,3 +487,49 @@ it('still flags MISSING_THROTTLE on sensitive routes with no rate-limiting', fun
     expect(issueTypes($analysis, 'route::POST::/login-noop'))
         ->toContain('MISSING_THROTTLE');
 });
+
+// =============================================================================
+// Custom auth guard detection via a verified `extends Authenticate` chain.
+// Catches guards whose class name does not contain "auth" (e.g. an
+// `auth.customer` alias mapped to App\Http\Middleware\AuthenticateCustomer),
+// which basename matching against the framework `Authenticate` alone misses.
+// =============================================================================
+
+function customAuthFixtureRoot(): string
+{
+    return fixture('custom-auth-project');
+}
+
+it('classifies a differently-named Authenticate subclass (resolved via alias) as authed', function () {
+    $route = makeSecurityRoute('DELETE', '/account/claims/{claim}', ['auth.customer']);
+    $registry = new MiddlewareRegistry([], [], [
+        'auth.customer' => 'App\\Http\\Middleware\\AuthenticateCustomer',
+    ]);
+
+    $analysis = (new SecurityAnalyzer)->analyze([$route], $registry, [], customAuthFixtureRoot());
+
+    $routeId = 'route::DELETE::/account/claims/{claim}';
+    expect($analysis[$routeId]['exposure'])->toBe('authed')
+        ->and(issueTypes($analysis, $routeId))->not->toContain('PUBLIC_WRITE');
+});
+
+it('classifies a differently-named Authenticate subclass given by bare FQCN as authed', function () {
+    $route = makeSecurityRoute('POST', '/account/profile', ['App\\Http\\Middleware\\AuthenticateCustomer']);
+    $registry = new MiddlewareRegistry([], [], []);
+
+    $analysis = (new SecurityAnalyzer)->analyze([$route], $registry, [], customAuthFixtureRoot());
+
+    expect($analysis['route::POST::/account/profile']['exposure'])->toBe('authed');
+});
+
+it('terminates on a cyclic middleware extends chain instead of exhausting memory', function () {
+    // CyclicA extends CyclicB extends CyclicA. Neither reaches Authenticate, so
+    // the route stays public; the point is that the walk returns at all.
+    $route = makeSecurityRoute('POST', '/cyclic', ['App\\Http\\Middleware\\CyclicA']);
+    $registry = new MiddlewareRegistry([], [], []);
+
+    $analysis = (new SecurityAnalyzer)->analyze([$route], $registry, [], customAuthFixtureRoot());
+
+    expect($analysis['route::POST::/cyclic']['exposure'])->toBe('public')
+        ->and(issueTypes($analysis, 'route::POST::/cyclic'))->toContain('PUBLIC_WRITE');
+});
--- a/tests/fixtures/custom-auth-project/app/Http/Middleware/AuthenticateCustomer.php
+++ b/tests/fixtures/custom-auth-project/app/Http/Middleware/AuthenticateCustomer.php
@@ -1,0 +1,14 @@
+<?php
+
+namespace App\Http\Middleware;
+
+use Illuminate\Auth\Middleware\Authenticate;
+
+/**
+ * Custom auth middleware backing a second guard (e.g. the `auth.customer`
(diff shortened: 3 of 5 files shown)


mrmarchone:
Thanks for your support, could you please change the pr to main branch?

tecnomanu:
Absolutely — already done! Rebased onto `main` and force-pushed. This PR composes cleanly with #49: it adds three things that don't overlap with the `extraAuthPatterns` mechanism — array-form alias parsing in `MiddlewareAnalyzer`, `extends Authenticate` chain detection in `SecurityAnalyzer` (so apps don't have to enumerate every custom guard in config), and a conservative-default `hasUnrecognisedMiddleware` check so unknown middleware no longer auto-fires PUBLIC_WRITE. My 6 tests were appended to the `SecurityAnalyzerTest.php` you created.

Full suite: 82/82 passing, phpstan clean, pint clean. Let me know if you'd like any adjustments :)

mrmarchone:
> Absolutely — already done! Rebased onto `main` and force-pushed. This PR composes cleanly with #49: it adds three things that don't overlap with the `extraAuthPatterns` mechanism — array-form alias parsing in `MiddlewareAnalyzer`, `extends Authenticate` chain detection in `SecurityAnalyzer` (so apps don't have to enumerate every custom guard in config), and a conservative-default `hasUnrecognisedMiddleware` check so unknown middleware no longer auto-fires PUBLIC_WRITE. My 6 tests were appended to the `SecurityAnalyzerTest.php` you created.
> 
> Full suite: 82/82 passing, phpstan clean, pint clean. Let me know if you'd like any adjustments :)

Could you please fix the conflicts , and if you could see this [PR](https://github.com/laramint/laravel-brain/pull/49) it has also the auth middleware please check it .

tecnomanu:
Done — rebased onto the current `main` and force-pushed. Now that #50 is merged the conflict was just a textual one in `SecurityAnalyzer.php` and `SecurityAnalyzerTest.php` (both PRs touch the same files); the changes compose cleanly. The whole PR sits on top of the new constructor signature (`extraAuthPatterns` + `extraThrottlePatterns` + `trustedRouteNames` + `trustedRouteUris`) and the pre-computed `$this->authPatterns` from #49 / #50.

Regarding the `extraAuthPatterns` mechanism from #49 — yep, that's exactly the infra this PR builds on. The three changes here are intentionally *complementary* to it:

1. **MiddlewareAnalyzer (array form)** — apps that register their custom guard via `$middleware->alias(['auth.customer' => AuthenticateCustomer::class])` (the form Laravel's docs and the bootstrap skeleton use) get their alias resolved into `MiddlewareRegistry` so the rest of the analyzer can match it. Before this, the array form was silently dropped and only the rarer 2-arg form worked.
2. **`extends Authenticate` chain detection** — `#49`'s `extraAuthPatterns` requires the user to enumerate every custom guard in config. This PR detects them automatically when the FQCN's `extends` chain reaches `Illuminate\\Auth\\Middleware\\Authenticate` (the standard multi-guard pattern). `extraAuthPatterns` still works for cases that don't follow that idiom — Bearer tokens, HMAC middleware, etc.
3. **Conservative `hasUnrecognisedMiddleware`** — when a route has middleware Brain doesn't recognise, we now keep `exposure='public'` (for visualisation) but suppress `PUBLIC_WRITE` because we can't *prove* the route is unauthenticated. Routes with no middleware or only safe-infra middleware (`web`, `throttle`, `bindings`, …) still fire `PUBLIC_WRITE` as before.

Full suite: **98 passing**, phpstan clean, pint clean. Let me know if you'd like anything tweaked :)

SanderMuller:
Ran this against current `main` (`e88edf1`) and three real Laravel applications. The diagnosis is
right, and part 1 is a clean fix. Two problems, both in the classification changes.

**`isAuthClass()` can recurse until memory runs out.** The memo is written after the recursive
call, so two middleware classes that extend each other make a plain `SecurityAnalyzer::analyze()`
exhaust memory and take the scan down with it. That needs a broken class to trigger, but a class
Brain can't execute shouldn't be able to kill a scan. Seeding the memo first fixes it:

```php
$this->authClassCache[$fqcn] = false;   // seed before recursing

$parent = $this->resolveParentClass($fqcn);
if ($parent === null) {
    return false;
}

return $this->authClassCache[$fqcn] = $this->isAuthClass($parent);
```

Verified: the cycle terminates and all 180 tests still pass.

**`hasUnrecognisedMiddleware()` suppresses more than unrecognised auth.**

| app | `main` | this PR |
|---|---:|---:|
| A | 32 | 25 |
| B | 53 | 29 |
| C | 12 | 3 |

App C is your case exactly: all nine suppressed routes carry a real custom auth middleware, and
`main` was wrong about every one. But of the 40 suppressions across the three, 23 are on routes
with no authenticating middleware at all. 17 are public AJAX endpoints whose only middleware is an
origin check, and 6 are hash-gated form endpoints whose only middleware is a performance monitor.
Both are single-middleware routes, so nothing else accounts for it. No `DELETE` was suppressed
anywhere, so the high-severity findings survived.

Using `isAuthClass()` inside `hasUnrecognisedMiddleware()` would keep C's fix and leave those 23
flagged. If that's too narrow, downgrade the finding or give it its own type. Dropping it silently
is the part I'd push back on.

Two minor ones: `locateMiddlewareClassFile()` duplicates `locateClassFile()` in the same class,
and `SAFE_INFRA_PATTERNS` lists `signed` and `ValidateSignature` as non-authenticating while
`AUTH_PATTERNS` deliberately includes both.

Part 1 stands on its own. Splitting it out would let it land now.

tecnomanu:
Thanks @SanderMuller — running it against three real apps is exactly the signal this needed. Pushed `84971b1` addressing all of it.

**1. `isAuthClass()` OOM on a cyclic extends chain — fixed.** Seeded the memo with `false` before recursing, as you suggested, so a cycle terminates instead of taking the scan down. Added a regression test (`CyclicA extends CyclicB extends CyclicA`) that asserts the walk returns and the route stays `public`.

**2. Over-broad suppression — removed `hasUnrecognisedMiddleware()` entirely rather than narrowing it.** Digging into your data made me realise the guard was redundant *and* wrong: `classifyExposure()` already runs `isAuthClass()` on every middleware, so any real Authenticate-subclass (registered alias or bare FQCN) is `authed` before the `PUBLIC_WRITE` gate is ever reached. That's what fixes App C — and it does so without the guard. The guard only ever fired on middleware that *didn't* resolve to auth, i.e. exactly your 23 genuinely-public routes (origin checks, perf monitors). So a mutating route that reaches the gate as `public` is now flagged again instead of dropped. This keeps C's nine suppressed and re-flags the 23. No `DELETE` behaviour changes.

I went with plain re-flagging rather than a downgraded/new finding type — in your three apps the unknowns were all genuinely public, so `PUBLIC_WRITE` is the correct call, and I'd rather not add a finding type on speculation. If you'd prefer a distinct lower-severity type for "mutating route behind middleware Brain couldn't resolve," happy to add it as a follow-up.

**3. `signed`/`ValidateSignature` contradiction — gone for free.** It only existed inside `SAFE_INFRA_PATTERNS`, which went away with the guard. `AUTH_PATTERNS` remains the single source of truth (signed URL = request auth).

**4. `locateMiddlewareClassFile()` duplication — folded into `locateClassFile()`.** One helper now: reflection first (so it works without a `projectRoot`), then the PSR-4 fallback. `resolveParentClass()` calls it.

Full suite green (99 passed), phpstan 0 errors, pint clean.

On splitting part 1 out to land now: glad to do that if you'd like the array-form alias fix in immediately — I can open a small PR with just the `MiddlewareAnalyzer` change and rebase this one on top. Your call.

tecnomanu:
Split done: **#86** carries just the `MiddlewareAnalyzer` array-form alias fix (Part 1), branched off current `main`, so it can land independently of the classifier discussion here.

Heads-up on this PR: its branch is based on `main` from before ~54 commits of security work landed (the basename-match `classifyExposure`, nested-group handling, etc.). GitHub still shows it as mergeable, but the `SecurityAnalyzer` here predates that rewrite. Rather than merge across that gap, I'd rebase Part 2 onto current `main` and reconcile it with the new basename matching before it's considered — the extends-chain check still adds value (it's a *verified* subclass check, so it catches differently-named guards like `AuthenticateCustomer` that basename matching alone misses), but it should be layered on top of what's there now, not the old version. Happy to do that rebase once #86 is in, or sooner if you'd prefer.

SanderMuller:
Re-ran the three-app census against `84971b1`. All four fixes verify, and I found one thing that changes what you should do next.

## The fixes check out

**1. Cyclic OOM.** Mutation-checked rather than taken on trust: with the seed line removed, `it terminates on a cyclic middleware extends chain` exhausts a 256 MB limit and dies inside `isAuthClass()`; with it in place the test passes. The test bites.

**2. Over-broad suppression — fully corrected.** Against your merge-base (`28e5720`), across three applications:

| app | PUBLIC_WRITE before | after | findings removed | removed **without** any `Authenticate`/`auth` middleware |
|---|---:|---:|---:|---:|
| A | 157 | 157 | 0 | 0 |
| B | 153 | 59 | 94 | **0** |
| C | 126 | 13 | 113 | **0** |

207 findings removed, every one on a route carrying a real auth middleware, none without. Nothing gained. `MISSING_THROTTLE` and `XSS_BLADE_UNESCAPED` are untouched on all three, so the change stayed inside the exposure gate.

Your claim that dropping the guard re-flags the genuinely-public routes holds, and it is bigger than either of us said. Measured against your own intermediate commit `d7bf3a3`, the guard was suppressing **134 / 29 / 10** routes on A / B / C that `84971b1` puts back.

**3 and 4.** `hasUnrecognisedMiddleware` and `SAFE_INFRA_PATTERNS` are gone from `src/`, and `locateClassFile()` is the single helper. Confirmed.

## The thing worth knowing before you go further

Your branch is **54 commits behind `main`**, and one of them is `b1d9094`, "Recognise app auth middleware subclasses so nested-group routes aren't flagged public" (fixes #59). It targets the same problem as your `SecurityAnalyzer` half. Same three apps, same census:

| app | your base | **current `main`** | this PR |
|---|---|---|---|
| A | PW=157, authed=1 | **PW=30, authed=358** | PW=157, authed=1 |
| B | PW=153, authed=10 | **PW=54, authed=188** | PW=59, authed=188 |
| C | PW=126, authed=0 | **PW=12, authed=300** | PW=13, authed=300 |

On A this PR changes nothing at all, because its base predates `b1d9094`; on B and C `main` already lands within a handful of findings of it. That is not a criticism of the work — it is that the branch is measured against a base where the problem was still open. Rebasing and re-measuring is the only way to see what your `SecurityAnalyzer` changes still add on top. It also explains the suite size: 99 tests on your base against 211 on `main`.

## So: yes, split part 1 out

Part 1 is **not** superseded, and I checked rather than assumed. `main` still does not register array-form aliases at all:

```php
$middleware->alias([
    'customer' => App\Http\Middleware\AuthenticateCustomer::class,
]);
```

```
current main:  {"customer":"(not registered)"}
this PR:       {"customer":"App\\Http\\Middleware\\AuthenticateCustomer"}
```

`main` has no `MiddlewareAnalyzer` changes since your base, and that array form is what Laravel's own skeleton and docs use — so every app registering aliases that way currently has them invisible to Brain. That is a clean, self-contained gap worth landing on its own.

My suggestion: open the small `MiddlewareAnalyzer` PR now, then rebase this one onto `main` and re-measure the `SecurityAnalyzer` half. If `b1d9094` already covers it, you have lost nothing — the cyclic-extends guard and its regression test are still worth keeping either way, since `b1d9094` did not add one.

Method, if you want to reproduce it: dump per-route exposure, issue types and resolved middleware to JSON on both arms and diff them. Counting findings is not enough — a finding dropped from a route carrying a real auth middleware is a false positive fixed, and the same finding dropped from a route carrying only a logger is a true positive lost. The middleware column is what tells them apart.

tecnomanu:
Rebased onto current `main` and force-pushed (`ebef7e1`). Two things changed since your review:

- The base moved ~54 commits, and `main` had already landed the basename-match `classifyExposure`, the current PUBLIC_WRITE gate, and dropped anything like the old `SAFE_INFRA_PATTERNS` suppression. So the over-broad suppression you pushed back on, the duplicated locate helper, and the `signed`/`ValidateSignature` contradiction are all gone with the rebase — nothing left to reconcile there.
- What remains is only the piece that's still missing on `main`: basename matching recognises a subclass literally named `Authenticate`, but not a differently-named guard like `AuthenticateCustomer`. This PR adds the verified extends-chain walk for exactly that case, with the memo seeded before recursing (your OOM fix) and a cyclic-chain regression test.

Net diff is now +182/−1, all on `SecurityAnalyzer` + tests. Suite green (213), phpstan 0, pint clean. Title/description updated to match the narrowed scope.

SanderMuller:
Re-measured `ebef7e1a` as asked. It rebases cleanly onto current `main` (0 behind), the `MiddlewareAnalyzer` half is correctly gone to #86, the cyclic-extends seed survived the rebase, and it is `213 passed`, PHPStan clean, Pint clean.

## The re-measurement you asked for: no change on three production apps

Against current `main`, same census as before:

| app | current `main` | this PR | routes moved |
|---|---|---|---:|
| A | PW=30, authed=358 | PW=30, authed=358 | 0 |
| B | PW=54, authed=188 | PW=54, authed=188 | 0 |
| C | PW=12, authed=300 | PW=12, authed=300 | 0 |

Identical everywhere. The reason is specific rather than damning: all three of those applications name their custom guard literally `App\Http\Middleware\Authenticate`, and `b1d9094` already matches that by name. My corpora simply cannot exercise what this PR adds.

## It is not redundant, and here is the proof

Two of *your own* tests fail on current `main`:

```
⨯ it classifies a differently-named Authenticate subclass (resolved via alias) as authed
⨯ it classifies a differently-named Authenticate subclass given by bare FQCN as authed
✓ it terminates on a cyclic middleware extends chain instead of exhausting memory
```

So `b1d9094` recognises a subclass whose *name* matches; this PR recognises one by walking the actual `extends` chain, whatever it is called. `AuthenticateCustomer`, `ShopGuard`, `TenantOnly` — none of those are caught today.

## The part worth knowing before merging either: they only work together

I built the shape a Laravel 11 skeleton actually generates — array-form alias map pointing at a differently-named guard, on a mutating route — and ran it on four arms:

```php
$middleware->alias(['auth.customer' => AuthenticateCustomer::class]);   // DELETE /account/claims/{claim}
```

| arm | alias registered | exposure | findings |
|---|---|---|---|
| current `main` | no | `public` | `PUBLIC_WRITE` ← false positive |
| #86 alone | yes | `public` | `PUBLIC_WRITE` ← still wrong |
| #51 alone | no | `public` | `PUBLIC_WRITE` ← still wrong |
| **#86 + #51** | yes | **`authed`** | **none** |

Neither PR fixes this on its own. #86 gets the alias into the registry; #51 recognises what it points at. Merged together they compose cleanly — I merged both onto `main` with no conflicts and got `215 passed`.

That also settles the redundancy question from my last comment: the `SecurityAnalyzer` half is *not* superseded by `b1d9094`. I was wrong to suspect it might be, and the three-app census alone would have kept me wrong — it took your own failing tests on `main` to show it.

## Two notes

Your cyclic-extends test **passes on current `main`**, so that guard is protecting the recursion this PR introduces rather than fixing a pre-existing bug. That is the right place for it — just flagging that it is not a fix `main` needs on its own, in case that affects how you describe it.

#86 still has the one-line issue from my review there: `$node->args[0]->value` on a single-argument call hits `VariadicPlaceholder`, which under Laravel's warning-to-exception handler kills the scan. Since these two now need to land together, that one is worth clearing first.

SanderMuller:
Now that #86 has merged, this one is the remaining half — re-confirmed against current `main`:

```
Laravel 11 shape: $middleware->alias(['auth.customer' => AuthenticateCustomer::class]) on a DELETE route

main (with #86)   alias registered: yes   exposure: public   issues: ["PUBLIC_WRITE"]   <- false positive
main + #51        alias registered: yes   exposure: authed   issues: []
```

So #86 got the alias into the registry and this closes it. 14/14 CI green and mergeable; it is 12 commits behind `main` but merges cleanly — I test-merged it onto current `main` and the combined tree builds and passes.

Nothing further from me on this one.

PR #52, opened 2026-06-22
Detect jobs dispatched via the Bus facade

SanderMuller:
`MethodTracer` already maps `Job::dispatch()` and `Event::dispatch()`, but the `Bus` facade isn't recognised — so jobs sent through `Bus::dispatch()`, `Bus::chain([...])`, or `Bus::batch([...])` don't appear in the graph.

This adds the `Bus` facade alongside the existing `Event` one, plus the matching label in `FlowExtractor`:

- `Bus::dispatch(new Job)` / `Bus::dispatchSync(new Job)` → one job edge
- `Bus::chain([new A, new B])` / `Bus::batch([new A, new B])` → a job edge per job in the array

It emits the same hop shape as the existing job detection (`'type' => 'job'`, `'method' => 'handle'`), so it flows straight into the current job-node / `dispatches`-edge handling with no downstream changes.

Like the existing dispatch detectors, only literal `new Job()` arguments are resolved — a job held in a variable isn't.

Tests and a small fixture are included; the full suite (Pint, Pest, PHPStan) is green.

### DIFF
--- a/src/Analysis/FlowExtractor.php
+++ b/src/Analysis/FlowExtractor.php
@@ -312,6 +312,9 @@ class FlowExtractor
             if (in_array($class, ['Event', 'Illuminate\\Support\\Facades\\Event']) && $method === 'dispatch') {
                 return ['type' => 'event', 'label' => 'Event::dispatch(...)'];
             }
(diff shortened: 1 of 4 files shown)


mrmarchone:
Thanks for you support, could you please change the pr to main branch?!

SanderMuller:
Done — retargeted both this PR and #53 to `main` (rebased onto the latest `main`, suite green). Thanks!

PR #57, opened 2026-06-28
feat: make route prefix configurable via config/env

vipmaa1:
Replace hardcoded `_laravel-brain` route prefix with `config('laravel-brain.route_prefix')`, backed by the `LARAVEL_BRAIN_ROUTE_PREFIX` env variable. Default unchanged.


vipmaa1:
# Why a configurable route prefix?

The default _laravel-brain prefix works for most setups, but breaks in deployments where the frontend and backend share the same domain and are split by URL path — `e.g. abcd.com` → frontend, `abcd.com/backend` → backend. In that case the backend's router only sees paths under /backend/..., so _laravel-brain routes become unreachable.

With this change you can align the prefix to your routing topology:

```LARAVEL_BRAIN_ROUTE_PREFIX=backend/_laravel-brain```

## Other practical benefits:

- Reverse proxy rules — match the prefix to whatever your nginx/Caddy/ALB rewrites to.
- Conflict avoidance — if another package or app already uses _laravel-brain, pick something else without touching package internals.
- Multi-tenant / sub-path installs — scope the UI under a project-specific path.

Default is unchanged (_laravel-brain), so existing installs need no action.

mrmarchone (changes requested):
Minor cosmetic nit (not blocking): ScanCommand.php:214 still prints the hardcoded /_laravel-brain "Open the viewer" URL, so the message will be wrong if someone overrides the prefix.

mrmarchone (changes requested):
🔴 Request Changes
Thanks for tackling this — the server-side wiring is clean and correct. However, this PR only makes the prefix configurable on the backend; the compiled frontend still hardcodes /_laravel-brain/, so setting LARAVEL_BRAIN_ROUTE_PREFIX to anything other than the default will break the viewer (all API calls and the logo will 404). It needs the frontend side before it can be merged as "configurable."

✅ What's correct
config/laravel-brain.php — new route_prefix key with LARAVEL_BRAIN_ROUTE_PREFIX env, default unchanged. 👍
routes/brain.php — group prefix now reads from config. 👍
src/Commands/ScanCommand.php — "Open viewer" URL derives from config. 👍
resources/views/index.blade.php — entry <script>/<link> tags templated via $brainPrefix. 👍
Test updated to use the config value. 👍
🔴 Blocking: the React SPA still hardcodes the prefix
The PR doesn't touch any frontend file, but the bundle bakes the prefix in at build time. With LARAVEL_BRAIN_ROUTE_PREFIX=custom:

frontend/vite.config.ts — base: '/_laravel-brain/' bakes import.meta.env.BASE_URL = '/_laravel-brain/' into the build. Every API call relies on it and will resolve to /_laravel-brain/api/... → 404:
App.tsx (api/scan), Toolbar.tsx (api/scan), AiRulesModal.tsx (api/generate-rules), Sidebar.tsx (api/context), hooks/useManifest.ts, hooks/useTabGraph.ts, hooks/useGraph.ts, hooks/useFileSource.ts (api/source).
Hardcoded string literals (not even parameterized by BASE_URL):
frontend/src/App.tsx:201 — 
frontend/src/components/StressTestPanel.tsx:180 — fetch(\/_laravel-brain/api/stress-test/${jobId}`)`
frontend/src/components/StressTestPanel.tsx:297 — fetch('/_laravel-brain/api/stress-test', …)
(There's already a Todo 1 comment at StressTestPanel.tsx:122-123 flagging this exact base-URL problem.)
Net effect: the blade loads the SPA from /custom/assets/... correctly, but once it boots, every request still targets /_laravel-brain/....

Requested changes
1. Propagate the runtime prefix from blade into the page. The SPA must read the prefix at runtime, not from the build-time base. In index.blade.php:

<script>window.__BRAIN_BASE__ = @json($brainPrefix . '/');</script>
(or a <base href="{{ $brainPrefix }}/"> tag if you prefer).

2. Centralize the base in the frontend. Add a single helper, e.g. frontend/src/lib/base.ts:

export const BASE = (window.__BRAIN_BASE__ ?? import.meta.env.BASE_URL) as string
Then replace every import.meta.env.BASE_URL usage and the three hardcoded /_laravel-brain/... literals (App.tsx logo, both StressTestPanel.tsx fetches) with BASE. This also resolves the existing Todo 1.

3. Rebuild the assets so the shipped bundle picks up the change. The committed filenames in index.blade.php (index-Ce-6gnQZ.js, etc.) will change — please regenerate them.

Note: leaving vite.config.ts's base: '/_laravel-brain/' is fine as a build-time default — it only matters that runtime code prefers window.__BRAIN_BASE__.

🟡 Minor
Consider documenting LARAVEL_BRAIN_ROUTE_PREFIX in the README alongside the other env vars.
Worth a test asserting the viewer works under a non-default prefix (even just that routes register and assets resolve) so this can't silently regress.

vipmaa1:
Fixed

mrmarchone:
Pipeline failed @vipmaa1

mrmarchone:
Unresolved comments

PR #61, opened 2026-07-04
Map the Blade view-composition tree as view → view edges

SanderMuller:
## What

Adds a `BladeViewAnalyzer` that maps the **view-composition tree** — which Blade view renders which other view — and wires the missing `view-to-view` ("renders") edges into the graph.

Brain already anchors a view to its rendering controller (`action-to-view`) but never descends into the views that view *itself* renders. Nested components and includes were captured only as inert node metadata (`parseBladeRefs`), so a change to a shared partial (e.g. an action-buttons component) reached no entry point and read as "no impact".

## What's discovered

Every reference form Blade offers:

- `@include` / `@includeIf` / **`@includeWhen`** / **`@includeUnless`** / **`@includeFirst`** / `@extends` / `@component` / `@each`
- modern **`<x-...>` component tags** (`<x-foo.bar>` → `components/foo/bar.blade.php`, with the folder-`index` form as a fallback)

Only references that resolve to a **real Blade file** under `resources/views` are linked, so a class-only `<x-...>` with no template, a namespaced `pkg::view`, or a dynamic `$var` name adds no edge.

## Route-anchored by design

`GraphBuilder::addViewComposition()` seeds the walk **only from view nodes already reached by an entry point** (route, console, channel, or Filament page — the step runs last, after all of those are wired), then descends transitively. So every nested component reachable from a rendered view is linked, without adding orphan views that nothing renders. Node ids are computed identically to the ones Brain already mints for a rendered view (`viewNodeId('blade::'+dot)`), so the new edges join the existing graph rather than creating duplicates. The walk is cycle-safe (a view including itself, or an A↔B loop, terminates).

## Tests

10 Pest tests in `BladeViewAnalyzerTest`: each directive + `<x-...>` resolution, folder-`index` fallback, the missing-file / namespaced / dynamic negatives, the pure `referencedViewNames` parser, transitive descent, **cycle termination**, **multi-seed shared-child dedup**, `@each`/`@includeWhen`/`@includeFirst`, `@include`+`<x->` dedup, and orphan exclusion.

Verified: `pint` clean · `phpstan` (level 5 + larastan) **No errors** · full suite **121 passed**. Dogfooded against a real ~large Laravel app: **205 composing views → 608 parent→child links** (`@extends` master templates, `@include` partials, and `<x-...>` components all resolved).

## Note

Independent of #58 (observers) and #60 (policies); all branch from `main` and touch adjacent spots in `GraphBuilder`/`ProjectAnalyzer`, so whichever merges later may need a trivial rebase. Happy to do that.


mrmarchone:
@SanderMuller Could you fix the conflicts ?

PR #79, opened 2026-08-10
Show the work a callback does in the method flow

### LINKED ISSUE #21: Method flow chart omits all data when using DB::transaction
Image

In my opinion transaction wrapped code blocks should be also analyzed in the Method Flow view.

P.S.
DB::beginTransaction and DB::commit would work just fine, however I find the DB::transaction call more comprehensive.

Image

SanderMuller:
Closes #21.

## What

A method whose body sits inside `DB::transaction(function () { ... })` charted as a single step —
the wrapper — and nothing else. Everything the transaction did was missing from the one view you
open to see what a method does.

`FlowExtractor` turned every call into a leaf and never looked at its arguments, so a closure
passed to one was never entered. Nothing about that is specific to transactions:
`Cache::remember`, `retry`, a collection `each`, anything taking a callback lost its contents the
same way.

```
before   DB::transaction(...)

after    DB::transaction(...)
           └─ $this->payer->charge()
              $this->ledger->record()
```

## The type matters more than the body

A call given a callback now carries that body, in the shape a `try` block already uses. It is also
retyped `loop`, which is what the viewer calls a block with a body — and that part is not cosmetic.
Both renderers in the shipped bundle read `body` for the `loop` type and for no other:

```js
else if (t.type === `loop`) { … t.body && t.body.length > 0 && walk(t.body, e) … }
```

A `call` carrying a `body` is dropped without a word. The fix would have looked correct in the
JSON, passed a test written against the JSON, and changed nothing on screen — which is the bug
being reported, not the fix for it. One of the tests pins the type for that reason.

Only the first callback of a call is entered. A call taking two is rare, and one body under one
label reads better than two merged.

## What it costs

Flow charts get bigger, so the question is whether they get unreadable. Measured over every action
node of two production applications:

| application | total flow steps | largest single chart | actions gaining a nested body |
|---|---|---|---|
| B — 1,885 files | 4,698 → 4,834 (+2.9%) | 65 → 65 | +38 |
| C — 4,152 files | 5,090 → 5,512 (+8.3%) | 50 → 54 | +81 |

The growth lands where work was hidden rather than spread across every chart, and the worst case
barely moves. Nothing else about a scan changes.

## Tests

`235 passed (806 assertions)` · PHPStan level 5 + larastan **No errors** · Pint clean.

Ten new, and the first tests `FlowExtractor` has had. Five fail on `main`: the transaction body,
the step type the viewer needs, an arrow function, a collection callback, and the N+1 that was
hidden inside one.

The other two pass on `main` and are there to fence the change in rather than to demonstrate it:
one asserts a call *without* a callback is untouched — no type change, no `body` key — and the
other that a callback merely containing a query is not itself an N+1. That second one exists
because I first read the new warnings this produces as false positives and drafted the PR saying
so. They are real: callback bodies are walked with `inLoop = false`, so a step can only be marked
by a loop genuinely nested inside.

## Bounded, after review

@mrmarchone found that following callbacks into each other had no depth limit, and that it
compounded with the pretty-print fallback in `shortExpr`: a call with one closure argument
rendered the whole closure into its label, so a chain re-rendered everything inside it at every
level. Reproduced — 4 ms at depth 40, 56 ms at 160, 514 ms at 320, and a fatal at 640 from
exhausting a 128 MB limit inside the pretty printer, which takes the scan with it.

A closure is now labelled by its signature (`function ($query, $user) {...}`), keeping the
parameters and dropping the body — which for a callback is charted underneath the step anyway.
And the descent stops at 32 levels, truncating rather than throwing: measured across three
production applications the deepest chart nests 5, out of ~14,000 steps, so that is about six
times real code. Depth 320 is now 2 ms and 640 is 1 ms.

That label change is visible in output: 452 labels change on one application and 414 on another,
and every changed line in a full dump is a `"label"` — no node, edge or step type moves.

### DIFF
--- a/src/Analysis/FlowExtractor.php
+++ b/src/Analysis/FlowExtractor.php
@@ -366,21 +366,43 @@ class FlowExtractor
      * @param  Node\Arg[]|Node\VariadicPlaceholder[]  $args
      * @return array<string, mixed>
      */
+    /**
+     * How deep this will follow callbacks into each other before it stops.
+     *
+     * Measured over three production applications: the deepest chart nests 5 levels, out of
+     * 14,000-odd steps, and only 25 steps anywhere reach level 4 or 5. So this is roughly six
+     * times what real code does, and it exists for source no one wrote by hand — FlowExtractor
+     * runs over whatever is in the project, and generated or pathological files should not be
+     * able to take a scan down.
+     */
+    private const MAX_CALLBACK_DEPTH = 32;
+
+    private int $callbackDepth = 0;
+
     private function withCallbackBody(array $step, array $args): array
     {
+        if ($this->callbackDepth >= self::MAX_CALLBACK_DEPTH) {
+            return $step;
+        }
+
         foreach ($args as $arg) {
             if (! $arg instanceof Node\Arg) {
                 continue;
             }
             $value = $arg->value;
-            if ($value instanceof Node\Expr\Closure) {
-                $body = $this->stmtsToSteps($value->stmts);
-            } elseif ($value instanceof Node\Expr\ArrowFunction) {
-                // An implicit return, matching how extractFromClosure() reads the same shape —
-                // otherwise one arrow function charts two ways depending on which path found it.
-                $body = $this->stmtsToSteps([new Node\Stmt\Return_($value->expr)]);
-            } else {
-                continue;
+            $this->callbackDepth++;
+            try {
+                if ($value instanceof Node\Expr\Closure) {
+                    $body = $this->stmtsToSteps($value->stmts);
+                } elseif ($value instanceof Node\Expr\ArrowFunction) {
+                    // An implicit return, matching how extractFromClosure() reads the same shape —
+                    // otherwise one arrow function charts two ways depending on which path found it.
+                    $body = $this->stmtsToSteps([new Node\Stmt\Return_($value->expr)]);
+                } else {
+                    continue;
+                }
+            } finally {
+                $this->callbackDepth--;
             }
 
             return $body === [] ? $step : ['type' => 'loop'] + $step + ['body' => $body];
@@ -479,12 +501,42 @@ class FlowExtractor
             return '!'.$this->shortExpr($expr->expr);
         }
 
+        // A closure or arrow function used as a value — nearly always the callback a step was
+        // built from. Without a case here it reaches the pretty printer below, which renders the
+        // entire body: for nested callbacks that means re-rendering everything inside this one,
+        // at every level, so the cost of labelling a chain grows with the square of its depth.
+        if ($expr instanceof Node\Expr\Closure) {
+            return 'function ('.$this->paramsLabel($expr->params).') {...}';
+        }
+        if ($expr instanceof Node\Expr\ArrowFunction) {
+            return 'fn ('.$this->paramsLabel($expr->params).') => ...';
+        }
+
         // Fallback: use pretty printer for full expression
         try {
             return $this->printer->prettyPrintExpr($expr);
         } catch (\Throwable) {
             return '...';
         }
+    }
+
+    /**
+     * The parameter names of a closure, which is the part of its signature worth keeping in a
+     * label. The body is what gets dropped: it is either charted underneath the step already, or
+     * it is an assignment whose contents belong in the code rather than in a chart label.
+     *
+     * @param  Node\Param[]  $params
+     */
+    private function paramsLabel(array $params): string
+    {
+        $names = [];
+        foreach ($params as $param) {
+            $names[] = $param->var instanceof Node\Expr\Variable && is_string($param->var->name)
+                ? '$'.$param->var->name
+                : '$?';
+        }
+
+        return implode(', ', $names);
     }
 
     private function argsLabel(array $args): string
--- a/src/Analysis/FlowExtractor.php
+++ b/src/Analysis/FlowExtractor.php
@@ -316,7 +316,7 @@ class FlowExtractor
                 return ['type' => 'dispatch', 'label' => "Bus::{$method}(...)"];
             }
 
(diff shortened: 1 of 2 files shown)


SanderMuller:
Marking draft — self-review found a regression this introduces.

Reusing the `loop` step type (so the viewer descends into `body`) also inherits its **N+1 semantics**. The detector flags a query inside a loop, so `DB::transaction(function () { ...query... })` is now reported as an N+1 that isn't one.

Measured on a 1,885-file application: 2 nodes gain a false `hasN1`. Ignoring `flowSteps`, those 2 nodes are the *only* graph difference — edges are untouched.

Fix direction: keep the `loop` type so the body still renders, but mark callback-derived blocks so the N+1 check skips them. A transaction is not a loop, and shouldn't be counted as one.

SanderMuller:
Correcting my previous comment — I had this backwards, and the evidence says the opposite.

The two nodes that gain `hasN1` are **true positives**, not false ones. Callback bodies are walked with `inLoop = false`, so they cannot mark themselves; the flag comes from a real `foreach` nested inside. One of the two:

```
chunkById(...)
  └─ foreach ($credentials as $credential)
       $payload = $credential->encrypted_credentials   ← n1
```

A genuine N+1 inside a `chunkById` callback that was never opened before. So this does not add false warnings — it surfaces real ones that were invisible, which is an argument for the change rather than against it.

Ignoring `flowSteps`, those 2 nodes are the only graph difference across a 1,885-file application; edges and manifests are untouched.

SanderMuller:
Correcting the Tests section: it went stale when I added two tests after writing it.

It said five new tests, 215 passed, four failing on `main`. Actually **seven new, 217 passed (767 assertions), five failing on main** — verified just now by running this branch's test file against a clean `origin/main` checkout. The fifth failure is the N+1 hidden inside a callback; the two that pass on main are the fences (a call without a callback stays untouched, and a callback merely containing a query is not itself an N+1) rather than demonstrations of the fix.

Nothing about the change itself moved — same commit, same behaviour, 14/14 green. Only the description was undercounting.

mrmarchone:
Hey @SanderMuller  — really like this change, making transaction/cache/retry/callback
bodies visible in the flow chart is a solid usability win, and the before/after
measurements on real apps are a nice touch.

One thing worth fixing before merge: withCallbackBody() (and the
stmtsToSteps() recursion it feeds, src/Analysis/FlowExtractor.php:363) has no
depth limit, and it compounds with the existing pretty-print fallback in
shortExpr — a closure without an explicit case there gets fully re-rendered as
text at every nesting level instead of once. Repro'd against synthetic
deeply-nested callback chains:

- Depth 320 → ~1.05s for a single file
- Depth 640 → memory-exhaustion fatal error (default 128MB limit), kills the
  whole scan
- Isolating pure recursion from the pretty-print cost, it segfaults around
  depth 15-16k — roughly half of what the pre-existing code path could handle
  before this change

Real hand-written app code won't nest callbacks that deep (matches your own
production numbers), but FlowExtractor runs over source we don't control —
one unusually generated or pathological file could take the whole scan down,
and right now nothing stops that.

Two small changes would close it:
1. Give shortExpr an explicit Closure/ArrowFunction case instead of falling
   through to the full pretty-printer.
2. Add a max-depth guard in withCallbackBody/stmtsToSteps that truncates
   gracefully (e.g. a placeholder step) past some reasonable nesting limit,
   instead of recursing unbounded.

Happy to pair on this or take a pass myself if useful — flagging before merge
since it's otherwise a clean, well-tested change.

mrmarchone (approved):

mrmarchone (changes requested):
RC @SanderMuller

SanderMuller:
Thanks @mrmarchone — reproduced it exactly as you described, and it was worse than a slow chart. Fixed in the commit above.

## Reproduction

Your numbers hold. A chain of calls each taking a closure:

| depth | before | after |
|---:|---:|---:|
| 40 | 4 ms | 1 ms |
| 160 | 56 ms | 1 ms |
| 320 | **514 ms** | 2 ms |
| 640 | **fatal — 128 MB exhausted inside the pretty printer** | 1 ms |
| 2,000 | — | 2 ms |

One correction to my own first attempt at reproducing it: with `DB::transaction()` I measured 2 ms at depth 320 and nearly concluded the report was wrong. That call has a dedicated handler with a short label, so it never reaches the fallback. It needs a call shape `shortExpr` has no case for — `$this->svc->wrap(function () { ... })` — and then it reproduces immediately. Worth knowing if anyone re-tests this.

## Both fixes, as you suggested

**`shortExpr` now has a closure case.** A closure is labelled by its signature: `function ($query, $user) {...}`, `fn ($item) => ...`. I kept the **parameters** rather than dropping the whole thing — they are the useful half of the signature and they are not what compounds; the body is what re-renders at every level.

**`withCallbackBody` stops at 32 levels**, truncating rather than throwing. I picked the number by measuring instead of guessing: across three production applications the deepest chart nests **5**, out of ~14,000 steps, and only 25 steps anywhere reach level 4 or 5. So 32 is about six times real code.

## What this changes in a real build

The label fix is visible, so it is worth stating plainly rather than burying: **452 labels change on one application and 414 on another**, and every single changed line in the full dump is a `"label"` — no node, edge, or step type moved.

```
- "label": "$category->questions->each(function (Question $question): void {\n    $question->allDependencies = $question->dependencies;\n})"
+ "label": "$category->questions->each(function ($question) {...})"
```

I think that is an improvement on its own terms — a multi-line source blob was never going to read well as a chart label, and for a callback the body is now charted underneath the step anyway, so the label was repeating it. But it is a change to output, not just to speed, and reviewers should see it as one.

Three tests, each checked by breaking the thing it guards: remove the depth limit and the deep-nesting test dies; drop the closure case and the label test fails.

Also rebased onto current `main` (it was 12 commits behind). `235 passed`, PHPStan clean, Pint clean.

PR #82, opened 2026-08-10
Let the rules extractor read source, not only a path

SanderMuller:
## What

`ValidationRulesExtractor` can only be reached with a path on disk. A caller that already holds
the source has to write a temp file to ask it anything.

The parser layer below it already offers both forms — `PhpFileParser::parseCode()` sits beside
`parse()` and returns the same shape. This adds the matching pair to the extractor:

```php
$extractor->hasNonAbstractRulesMethodInCode($source);
$extractor->extractFromCode($source);
```

Three kinds of caller are shut out today: something reading a git blob, which never has a path;
an editor or language-server integration looking at an unsaved buffer; and anything that has
already read the file and would otherwise pay for a second read and a second parse.

## The file variant does not go through the code variant

The obvious refactor is to have `extractFromFile()` read the file and hand the string to
`extractFromCode()`. That would be wrong here. `PhpFileParser::parse()` is what consults the
shared parse cache; `parseCode()` always parses. `hasNonAbstractRulesMethod()` is asked thousands
of times in a build, so routing paths through the code variant re-reads and re-parses every file
on every call.

That is not a guess. Built both ways and measured a full scan:

| application | meeting after the parse | file variant delegating to the code variant |
|---|---:|---:|
| B | 1,818 ms | 5,381 ms — **3.0×** |
| A | 2,667 ms | 13,067 ms — **4.9×** |

Medians with the arms interleaved back to back, three reps on B and two on A. The spread within
each arm is under 10%, so nothing here rests on the rep count.

So the two entry points meet *after* parsing instead, at a private method taking the AST. There
is still exactly one implementation of the analysis, which is the part worth not duplicating.

Two things are worth knowing about that measurement, because both would let this regression ship
unnoticed. The test suite passes either way — a parity test compares the two variants to each
other, and both are equally correct when one is four times slower. And the parse counter reads
1,582 either way, because `PhpFileParser::$parseCount` is incremented in `parseFile()`, not in
`parseCode()`, so parses driven through the code path are invisible to it. Wall clock is the only
instrument that sees this.

## What this is not

It does not make the extractor usable for comparing two revisions of a form request. That needs
something this return shape cannot express: the ability to say *"this side cannot be
enumerated"* and stay silent. An unkeyed item or a spread can hide any number of fields, a
class-constant key resolves against whichever codebase is loaded, and a `rules()` that builds its
array up before returning it is not readable at all. `list<RuleRow>` has no way to report that —
a keyless item comes back as the field `'*'`, which is a value, not a signal. Field names also
arrive as pretty-printed source, quotes included.

Serving that would mean a second, stricter contract, which is a different change and arguably a
different package's job. This one enumerates one present-tense file for a graph.

## Nothing else moves

byte-identical to `main` on three applications. Scan time and parse counts are unchanged against
`main` with the arms interleaved, which is what "additive" should mean.

## Tests

`230 passed (794 assertions)` · PHPStan level 5 + larastan **No errors** · Pint clean.

Five new. One asks the same fixture both ways and asserts the answers are equal, so a change that
reaches only one variant fails here. The rest cover what only the new entry point can reach:
source that never was a file, source that does not parse, an abstract `rules()` — the one branch
of the shared path a happy-path test never touches — and the difference between a `rules()` this
cannot read and no `rules()` at all, since both return no rows and only one of them means there
is no validation.

### DIFF
--- a/src/Analysis/ValidationRulesExtractor.php
+++ b/src/Analysis/ValidationRulesExtractor.php
@@ -51,6 +51,13 @@ final class ValidationRulesExtractor
         $this->printer = new PrettyPrinter;
     }
 
+    /**
+     * Asking about a path goes through the memo above and, under it,
+     * {@see PhpFileParser::parse()} and the parse cache it shares with every other analyzer. Do
+     * not reroute this through the code variant below: `parseCode()` has neither, so a path
+     * handed to it is re-read and re-parsed on every call, and this question is asked thousands
+     * of times in a build.
+     */
     public function hasNonAbstractRulesMethod(string $file): bool
     {
         $stat = @stat($file);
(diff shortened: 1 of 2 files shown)


SanderMuller:
Went back over this and checked the claims rather than restating them. One correction to the description, and the underlying design claim turned out to be understated.

**I had claimed the alternative shape would be slower without having measured it.** Now built both ways and measured a full scan: routing the file variants through the code variant is **3.0× slower on one application and 4.6× on another** (1,649 → 4,975 ms and 2,547 → 11,781 ms). Worse than I assumed, and now in the description as a measurement rather than an assertion.

**A sentence in the description was wrong.** It offered the unchanged parse count as evidence that the shape protects the cache. It is not evidence: `PhpFileParser::$parseCount` is incremented in `parseFile()`, not `parseCode()`, so the counter reads 1,582 under the fast shape *and* under the 3× slower one. The test suite passes under both too, since a parity test compares the two variants to each other and both stay correct. Wall clock is the only instrument that catches this, which is worth knowing for anyone who touches this seam later. Replaced with what the numbers actually show.

Byte-identical output re-verified against a clean `origin/main` checkout after the simplification commit, and scan time and parse counts are unchanged against `main` with the arms interleaved.

mrmarchone:
@SanderMuller Could you please fix the conflicts ?

PR #84, opened 2026-08-14
Refuse a scoped rebuild that cannot name what it claims to rebuild

### LINKED ISSUE #83: scopedTo(): a scope matching no provenance passes the soundness check and returns the stale graph
# laravel-brain: a scoped rebuild can return a stale graph instead of throwing

Observed against laravel-brain 2.4.0 while wiring `ProjectAnalyzer::scopedTo()` into a consumer.

## Summary

`scopedTo($files, $previous)` matches `$files` against the project **two different ways**, and the
mismatch between them turns the soundness check into a no-op for any path form that only one of them
accepts. The result is not an error — it is a previous graph handed back as though it were current.

| Site | How it matches `$files` |
|---|---|
| Controller filter, `ProjectAnalyzer.php:201-206` | `realpath()` on **both** sides |
| Soundness check, via `GraphProvenance::edgeIdsForFiles` | **verbatim** against each node's `data['file']`, no normalisation |

`GraphProvenance::of()` keys `byFile` on `$node->data['file']` exactly as the analysis recorded it —
the un-normalised absolute path derived from `$projectRoot`.

## Why it fails silently

`ProjectAnalyzer` guards the merge with:

```php
if (IncrementalMerge::ownedEdgeKeySet($mergeInto, $scopeToFiles)
    != IncrementalMerge::ownedEdgeKeySet($fullGraph, $scopeToFiles)) {
    throw new ScopedRebuildNotApplicable;
}
```

When `$scopeToFiles` matches nothing in the provenance, `edgeIdsForFiles()` returns `[]` for **both**
graphs. Two empty sets compare equal, the guard passes, and `applyPartial()` substitutes nothing.

So the failure mode is: *the less a scope matches, the more confidently the check approves it.*

## Reproduction

On macOS this needs nothing exotic — `realpath()` alone triggers it, because `/var` resolves to
`/private/var` while the provenance keeps the unresolved form. A caller doing the natural thing
(`realpath($changedFile)`) hits it every time under `sys_get_temp_dir()`.

Fixture: one route → one controller → `Publisher::run()` → `Publisher::log()`. Then edit
`Publisher.php` so the call moves (`run()` empty, `log()` calls `run()`) — an edit the check exists
to refuse — and scope to `realpath('…/Publisher.php')`:

```
PATH=scoped                     <- accepted; should have thrown
scoped nodes=5 edges=4          <- stale: still carries run→log
full   nodes=4 edges=3
```

Passing the un-realpath'd path instead makes the same edit throw `ScopedRebuildNotApplicable`
correctly.

## Suggested fix

Two options, not exclusive:

1. **Normalise both sides of the soundness check** the way the controller filter already does, so any
   caller-supplied path form works consistently across the feature.
2. **Treat a zero-match scope as inapplicable.** A scope naming files the previous graph attributes
   nothing to cannot be substituted, so `ScopedRebuildNotApplicable` is the honest answer rather than
   "the sets are equal". This is the more robust of the two: it fails closed regardless of why the
   match came up empty.

## Consumer-side mitigation already in place

The caller now requires every scoped path to be a key of
`GraphProvenance::of($previous)->byFile` and refuses the whole scope otherwise, passing paths in the
provenance's own form. That closes it for this consumer, but the trap stays open for any other
`scopedTo()` caller, which is why it is worth fixing upstream.

SanderMuller:
Closes #83.

## What

A scoped rebuild could hand back the previous graph as though it were current.

`scopedTo($files, $previous)` matched `$files` against the project two different ways. The
controller filter resolves both sides with `realpath()`; the soundness check matched them verbatim
against the path each node recorded. Any path form only one of them accepted fell between the two.

The check compares what the scope owns in the previous graph against what it owns in the fresh
one. A scope that matches nothing owns nothing in **both**, two empty sets compare equal, the
check approves, and the merge substitutes nothing. The less a scope matched, the more confidently
it was accepted.

Reproduced with a route → controller → service fixture, editing the controller so its call moves
from `Publisher::run()` to `Publisher::log()` — the edit the check exists to refuse:

```
scope as the build recorded it   ScopedRebuildNotApplicable   correct
scope through realpath()         accepted, and the graph still says index → Publisher::run
```

Both graphs have four nodes and three edges. The stale one is not smaller, it is wrong — so a
check that counts sees nothing, which is why the tests here compare edges by endpoint, type and
label.

## Two changes

**`GraphProvenance` resolves a caller's path form.** `byFile` is keyed on `data.file` as the
analysis wrote it. A lookup that misses verbatim is retried against the recorded keys resolved
through `realpath()`, so the ownership lookup agrees with the controller filter that already
worked this way. It is lazy and only runs after a miss, so the case that always hits — every path
watch mode passes, which a build produced itself — costs nothing.

**A scope that cannot name anything is refused up front.** Path form was the common way in, not
the only one: a path naming no file at all was accepted too, with the same result. Every scoped
path now has to be an existing file inside the project, checked before any work happens. A file
deleted since the previous run, a path from another checkout, a typo, and an empty scope are all
`ScopedRebuildNotApplicable` rather than a stale graph.

## What it deliberately does not do

The issue's second suggestion was to refuse any scope matching no provenance. That is sound, and
it is too blunt: a file the graph attributes nothing to is the common case, not the exception,
because the graph only holds what an entry point reaches.

| application | php files in app/ | graph attributes something | attributes nothing |
|---|---:|---:|---:|
| A | 1,089 | 596 (55%) | 493 (45%) |
| B | 1,899 | 840 (44%) | 1,059 (56%) |
| C | 4,181 | 597 (14%) | **3,584 (86%)** |

That rule would send 45–86% of single-file edits down the full-rebuild path, which is most of what
watch mode exists to avoid, and it buys little: an edit to a file nothing reaches cannot change
the graph without an edit to its caller, and that caller is a changed file too, is therefore in
the scope, and does own provenance. A test pins that this case still takes the scoped path.

## Scope of the bug

Watch mode was never affected. `ScanCommand` passes `base_path()` unmodified into both the mtime
walk and `analyze()`, and the directory iterator preserves that form, so its scope paths and the
recorded paths agree by construction. This reached consumers calling `scopedTo()` directly, which
is how it was found.

## Tests

`217 passed (759 assertions)` · PHPStan level 5 + larastan **No errors** · Pint clean.

Seven new, and the first tests `scopedTo()` has had — the classes it composes were each covered,
the seam that composes them was not, which is why this shipped. Four fail on `main`: the
normalised path, a path naming nothing, a path outside the project, and an empty scope.

The other three pass on `main` and are there to fence the change in: a scoped rebuild still
reproduces a full build when the edit left the call graph alone, still refuses an edit that moved
a call, and still takes the scoped path for a file the graph attributes nothing to.

The normalised-path test builds through a symlink it creates rather than relying on the platform's
temp directory being one — it is on macOS, where `/var` resolves to `/private/var`, and is not on
Linux, so without that the test would quietly assert nothing on CI. It asserts the two path forms
really do differ before relying on them.

## Gate

byte-identical to `main` on three applications, and scan time and parse counts are unchanged with
the arms interleaved.

### DIFF
--- a/src/Analysis/Incremental/GraphProvenance.php
+++ b/src/Analysis/Incremental/GraphProvenance.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace LaraMint\LaravelBrain\Analysis\Incremental;
 
+use LaraMint\LaravelBrain\Analysis\ProjectAnalyzer;
 use LaraMint\LaravelBrain\Graph\Graph;
 
 /**
@@ -34,6 +35,13 @@ use LaraMint\LaravelBrain\Graph\Graph;
  */
 final class GraphProvenance
 {
+    /**
+     * realpath() of each key in {@see $byFile} => that key, built on the first lookup that misses.
+     *
+     * @var array<string, string>|null
+     */
+    private ?array $resolvedKeys = null;
+
     /**
      * @param  array<string, array{nodes: string[], edges: string[]}>  $byFile
      * @param  array<string, string>  $nodeFile  nodeId => owning file
@@ -75,7 +83,7 @@ final class GraphProvenance
     {
         $ids = [];
         foreach ($files as $f) {
-            foreach ($this->byFile[$f]['nodes'] ?? [] as $id) {
+            foreach ($this->byFile[$this->recordedKey($f)]['nodes'] ?? [] as $id) {
                 $ids[] = $id;
             }
         }
@@ -88,11 +96,51 @@ final class GraphProvenance
     {
         $ids = [];
         foreach ($files as $f) {
-            foreach ($this->byFile[$f]['edges'] ?? [] as $id) {
+            foreach ($this->byFile[$this->recordedKey($f)]['edges'] ?? [] as $id) {
                 $ids[] = $id;
             }
         }
 
         return array_values(array_unique($ids));
+    }
+
+    /**
+     * The key the build recorded for this file, or the path unchanged when nothing matches it.
+     *
+     * `byFile` is keyed on `data.file` exactly as the analysis wrote it: the project root the
+     * caller passed, with a relative path appended. A caller that normalises its own paths first
+     * — `realpath()` is the obvious thing to reach for, and on macOS it rewrites `/var` to
+     * `/private/var` — would otherwise name the right file in a form nothing here recognises and
+     * silently own nothing at all. The controller filter in {@see ProjectAnalyzer::analyze()}
+     * already resolves both sides of its match; this makes the ownership lookup agree with it.
+     *
+     * Resolved lazily, and only after a verbatim lookup misses, so the case that always hits —
+     * the paths a build produced itself, which is every path watch mode passes — costs nothing.
+     */
+    private function recordedKey(string $file): string
+    {
+        if (isset($this->byFile[$file])) {
+            return $file;
+        }
+
+        $real = realpath($file);
+        if ($real === false) {
+            return $file;
+        }
+
+        if ($this->resolvedKeys === null) {
+            $this->resolvedKeys = [];
+            foreach (array_keys($this->byFile) as $known) {
+                if ($known === '') {
+                    continue;
+                }
+                $knownReal = realpath($known);
+                if ($knownReal !== false) {
+                    $this->resolvedKeys[$knownReal] = $known;
+                }
+            }
+        }
+
+        return $this->resolvedKeys[$real] ?? $file;
     }
 }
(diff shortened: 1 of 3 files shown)


SanderMuller:
**Heads-up on merge order.** #84 and #85 both add a private method immediately above `ProjectAnalyzer::analyze()`, so whichever lands second will show a conflict in `src/Analysis/ProjectAnalyzer.php`. It is purely positional — the two additions are independent, and the resolution is to keep both.

One wrinkle worth stating because I hit it: the conflict region opens *inside* a shared docblock, so a naive keep-both nests a method inside a comment and the file stops parsing. The junction needs the first docblock closed and a new one opened.

I test-merged them: resolved that way, the combined tree is **220 passed (763 assertions)**, PHPStan clean, Pint clean, with every test from both PRs passing. Happy to rebase whichever one lands second if that is easier.

PR #86, opened 2026-08-24
fix(middleware): read array-form $middleware->alias([...]) registrations

tecnomanu:
Split out of #51 as @SanderMuller suggested ("Part 1 stands on its own. Splitting it out would let it land now") — this is just the `MiddlewareAnalyzer` half, with none of the security-classifier changes still under discussion there.

## Problem

`MiddlewareAnalyzer::extractAliases()` only handled the two-argument form:

```php
$middleware->alias('key', Class::class);   // ✓ handled
```

The array form Laravel's docs and the `bootstrap/app.php` skeleton recommend was silently dropped:

```php
$middleware->alias([
    'auth.customer' => AuthenticateCustomer::class,   // ✗ never registered
]);
```

So any custom guard registered the array way never made it into the `MiddlewareRegistry`, and everything downstream that resolves aliases (exposure classification included) was working from an incomplete map.

## Change

`extractAliases()` now handles both forms:

- **Form A** — `$middleware->alias('key', Class::class)` — unchanged, still supported.
- **Form B** — `$middleware->alias(['key' => Class::class, ...])` — new. Iterates the array items, taking each string key as the alias and resolving the value through the existing `extractClassString()`.

No public API change; no new constructor parameters.

## Tests

New `tests/Unit/MiddlewareAliasFormTest.php` + a `middleware-array-alias` fixture whose `bootstrap/app.php` registers both forms in one file:

- extracts the two aliases from an array-form `$middleware->alias([...])` call
- still extracts the two-argument `$middleware->alias(key, class)` form

Full suite green (212 passed), phpstan 0 errors, pint clean.

### DIFF
--- a/src/Analysis/MiddlewareAnalyzer.php
+++ b/src/Analysis/MiddlewareAnalyzer.php
@@ -185,13 +185,33 @@ class MiddlewareAnalyzer
 
             private function extractAliases(Node\Expr\MethodCall $node): void
             {
-                if (count($node->args) >= 2) {
-                    $alias = $node->args[0]->value instanceof Node\Scalar\String_
-                        ? $node->args[0]->value->value
-                        : null;
+                if (count($node->args) === 0) {
+                    return;
+                }
+
+                // Form A: `$middleware->alias('key', Class::class)`
+                if (count($node->args) >= 2 && $node->args[0]->value instanceof Node\Scalar\String_) {
+                    $alias = $node->args[0]->value->value;
                     $class = $this->extractClassString($node->args[1]->value);
-                    if ($alias && $class) {
+                    if ($alias !== '' && $class !== null) {
                         $this->aliases[$alias] = $class;
+                    }
+
+                    return;
+                }
+
+                // Form B: `$middleware->alias(['key' => Class::class, ...])`
+                // This is the form Laravel's docs and the bootstrap/app.php
+                // skeleton use, so most real apps register custom aliases this way.
+                if ($node->args[0]->value instanceof Node\Expr\Array_) {
+                    foreach ($node->args[0]->value->items as $item) {
+                        if (! $item || ! ($item->key instanceof Node\Scalar\String_)) {
+                            continue;
+                        }
+                        $class = $this->extractClassString($item->value);
+                        if ($class !== null) {
+                            $this->aliases[$item->key->value] = $class;
+                        }
                     }
                 }
             }
--- a/src/Analysis/MiddlewareAnalyzer.php
+++ b/src/Analysis/MiddlewareAnalyzer.php
@@ -185,7 +185,13 @@ class MiddlewareAnalyzer
 
             private function extractAliases(Node\Expr\MethodCall $node): void
             {
-                if (count($node->args) === 0) {
+                // First-class callable syntax `$middleware->alias(...)` puts a
+                // VariadicPlaceholder (no `->value`) in args[0]. Reading it would
+                // raise a warning that Laravel's HandleExceptions turns into an
+                // ErrorException, killing the scan — so bail unless args[0] is a
+                // real Node\Arg. Matches the guard used across MethodTracer /
+                // FlowExtractor.
+                if (count($node->args) === 0 || ! $node->args[0] instanceof Node\Arg) {
                     return;
                 }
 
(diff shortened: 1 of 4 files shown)


SanderMuller:
Reviewed and measured. The change does what it says, it is rebased on current `main`, and the gap it closes is real — I re-confirmed that `main` returns `(not registered)` for the documented array form while this branch resolves it. Two findings, one of which I would fix before merging.

## One bug: this can now kill a scan

`extractAliases()` reaches `$node->args[0]->value` when there is only one argument. PHP's first-class callable syntax puts a `VariadicPlaceholder` there, which has no `->value`:

```php
$middleware->alias(...);   // legal PHP 8.1+
```

```
current main:  scan completed, aliases=[]
this branch:   Warning: Undefined property: PhpParser\Node\VariadicPlaceholder::$value
```

A warning looks harmless, but every `artisan` command runs with Laravel's `HandleExceptions` bootstrapper, which turns warnings into `ErrorException`. With that handler installed:

```
current main:  scan completed, aliases=[]
this branch:   SCAN DIED: ErrorException: Undefined property: PhpParser\Node\VariadicPlaceholder::$value
```

`main` is immune only by accident — it touches `args[0]` solely inside `count($node->args) >= 2`, and a callable placeholder gives one argument. The new early-out reaches it with one.

This is the same principle you and I landed on in #51: code Brain cannot make sense of should not be able to take the scan down. Fix is one line, using the guard this codebase already uses in six places (`MethodTracer.php:408, 620, 827, 837, 868`, `FlowExtractor.php:454`):

```php
if (count($node->args) === 0 || ! $node->args[0] instanceof Node\Arg) {
    return;
}
```

I applied exactly that and re-ran: the crash is gone, all six call shapes below still behave, and `212 passed` is unchanged.

## What I verified

Six shapes, `main` vs this branch:

| call | main | this branch |
|---|---|---|
| `alias(['k' => K::class])` | not registered | **`k => K`** |
| `alias('k2', K2::class)` | `k2 => K2` | `k2 => K2` |
| `alias(aliases: [...])` | not registered | **`k3 => K3`** |
| `alias($map)` | – | – |
| `alias(['k' => [K::class]])` | – | – |
| `alias(...)` | – | **warning / ErrorException** |

The named-argument row is a bonus your description does not claim — it works because `args[0]->value` is still the array. Worth a test so it does not regress silently.

Full build output — graph, subgraphs, manifest, totals, tracer edge list — is **byte-identical to `main`** on three production applications, so nothing regresses there.

## An honest limit on the evidence

I could not measure the benefit on those three, because none of them uses this form: all are legacy `Kernel.php` `$routeMiddleware` apps, with a `bootstrap/app.php` that never calls `withMiddleware()`. So my corpora prove *no regression* and cannot prove *value*. The value case rests on the synthetic probe plus the fact that this is what Laravel's own skeleton generates — which I think is enough, but it is not the same class of evidence as the #51 numbers, and I would rather say so than let the byte-identical result read as an endorsement of impact.

Two smaller notes:

- The Form A condition changed from `if ($alias && $class)` to `if ($alias !== '' && $class !== null)`. An alias literally named `"0"` now registers where it previously did not. Almost certainly an improvement, just flagging it as a real behaviour change.
- `['k' => [K::class]]` yields nothing on both arms. Fine — an alias value is a single class — just noting the shape is unsupported rather than silently half-handled.

With the `Node\Arg` guard added, this is good to go from my side.

SanderMuller:
Cross-reference: I re-measured the rebased #51 and found these two only work together. On the shape a Laravel 11 skeleton generates — `$middleware->alias(['auth.customer' => AuthenticateCustomer::class])` on a mutating route — #86 alone registers the alias but the route still classifies `public` with a false `PUBLIC_WRITE`; #51 alone never sees the alias at all; both together give `authed` with no finding. Details and the four-arm table are on #51.

So this one is not optional-nice-to-have, it is half of a fix. Which makes the `VariadicPlaceholder` guard above worth clearing before they land — it is one line, and the codebase already uses that exact guard in six places.

tecnomanu:
Fixed in `5be8586`. Applied the exact `Node\Arg` guard you pointed to:

```php
if (count($node->args) === 0 || ! $node->args[0] instanceof Node\Arg) {
    return;
}
```

Confirmed the fixture actually exercises both shapes — parsing it yields `Node\Arg` for the named-arg call and `VariadicPlaceholder` for `alias(...)`, so the guard is what stops the crashing property read.

Added two tests to lock it in:
- `alias(aliases: [...])` (named argument) registers — the bonus you spotted, now explicit so it can't regress silently.
- `alias(...)` (first-class callable) no longer crashes and registers nothing.

Full suite green (214 passed), phpstan 0, pint clean.

On your two smaller notes: the `"0"` alias behaviour change and the `['k' => [K::class]]` shape being unsupported are both intentional/acceptable as you said, so I left them as-is. And noted on the cross-reference — since #51 and this need to land together, this one's now clear from your side.

PR #88, opened 2026-08-27
Feat: modular filament support

webard:
## Filament: configurable search paths, panels outside the `*PanelProvider` convention, and base-class subclasses

`FilamentAnalyzer` currently hardcodes the default Laravel skeleton: panels are read from `app/Providers/Filament` and everything else from `app/Filament`, with no configuration surface at all. Two consequences:

1. **An application that keeps its Filament classes anywhere else gets a silently empty Filament graph.** The `vendor/filament/filament` check still reports `detected: true`, every scan then hits `is_dir()` on a path that does not exist, and the scan summary prints `0 resources` with no indication that nothing was even looked at. A modular monolith - local Composer packages under `app-modules/*`, no `app/` directory whatsoever - is the case I hit, but the same applies to any non-default layout.
2. **A project base class hides every resource.** `extractResource()` accepted only a direct `extends` whose last segment is `Resource`, and `extractRelationManager()` only a direct `extends RelationManager`. An `AppResource extends Resource` sitting between Filament and the real resources is an extremely common pattern, and it drops all of them.

### What this changes

**`filament.panel_paths` / `filament.paths` config keys.** Both default to today's values, so existing installs behave identically. An entry is used as-is when it is a directory and expanded as a glob pattern otherwise:

```php
// config/laravel-brain.php
'filament' => [
    'panel_paths' => ['app-modules/*/src/Filament'],
    'paths' => ['app-modules/*/src/Filament'],
],
```

**Panels are no longer required to be named `*PanelProvider.php`.** That filename still counts as a panel by convention - it is what Filament's own installer writes, and it configures the injected `$panel` rather than building one. Any other file counts only when it actually builds a `Panel::make()` chain (resolved through the file's use map, so an aliased import still matches). Without that guard, pointing `panel_paths` at a whole source tree would turn every class into a panel; with it, a panel registered from a plugin class or a factory is found.

**Resources and relation managers are recognised through their whole `extends` chain.** A `class FQCN => parent FQCN` map is collected once from the configured trees and walked when deciding what a class is. The walk terminates at a parent that was never parsed (a base class in `vendor/`), where the short name decides - exactly what the single-level check did before, so this is strictly additive.

Pages and widgets were already ungated and are untouched. Panel linking (`->resources([...])`, `->discoverResources(for: '...')`), slug/route computation and everything downstream are unchanged.

### Incidental simplifications

- `scanWidgets()` had two hand-rolled passes with a shared `$seenFiles` guard (one for `app/Filament/Widgets`, one for `app/Filament/Resources/**/Widgets/`). Both are now one pass filtered on a `/Widgets/` path segment, which covers the same two locations plus page-level widget directories. De-duplication moved into the shared file iterator.
- The four scan methods used to re-walk the tree independently; `analyze()` now resolves the directories once and passes them down.

### Tests

8 new tests in `tests/Unit/FilamentAnalyzerTest.php` against a new `tests/fixtures/filament-modular-project` fixture - two modules, a panel built by `Panel::make()` outside the naming convention, a resource and a relation manager behind project base classes, and two negative controls (a plain class inside a panel path, a plain class inside `Resources/`). They assert, among others, that the same fixture yields nothing under the default `app/` paths and everything under the configured ones.

Existing Filament tests are unchanged and still pass. Full suite: `262 passed (890 assertions)`, `pint` clean, `phpstan` 0 errors.

A ~90-module monolith with no `app/` directory, before → after:

| | before | after |
|---|---|---|
| panels | 0 | 1 |
| resources | 0 | 55 |
| pages | 0 | 217 |
| widgets | 0 | 87 |
| relation managers | 0 | 22 |

(0.8 s for the Filament pass over ~830 files.) With only the config keys and not the `extends`-chain fix, resources came out at 4 of 55 - the 51 behind the project base class.

### Compatibility

The defaults are the previous hardcoded paths, `*PanelProvider.php` still short-circuits panel detection, and the chain walk can only match more than the direct-parent check did. `FilamentAnalyzer::__construct()` gains two optional arguments.

**Two things do change on a default layout, both in the direction of finding more.** The single `/Widgets/` pass replaces two hand-rolled ones and picks up a widget under `app/Filament/<PanelName>/Widgets/`, which neither of the old passes looked in. And the `extends` chain now recognises a base class in the same namespace as its children — three relation managers on the application this was developed against were invisible before it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -265,6 +265,20 @@ This produces the full edge list used to build the graph.
 
 When Filament is installed, the scanner discovers every panel registered via service providers, then resolves its resources, pages, widgets, and relation managers — both explicitly listed (`->resources([...])`) and auto-discovered (`->discoverResources(for: '...')`). Filament page methods are traced through the same call-chain engine as controller actions, so models and services they touch appear in the graph.
 
+Resources and relation managers are recognised through their whole `extends` chain, so a project base class (`class OrderResource extends AppResource`, `AppResource extends Resource`) does not hide them.
+
+Applications that keep their Filament classes somewhere other than `app/Filament` - a modular monolith with no `app/` directory, for instance — point the scanner at their own layout. An entry is used as-is when it is a directory and expanded as a glob pattern otherwise:
+
+```php
+// config/laravel-brain.php
+'filament' => [
+    'panel_paths' => ['app-modules/*/src/Filament'],
+    'paths' => ['app-modules/*/src/Filament'],
+],
+```
+
+A file named `*PanelProvider.php` is treated as a panel by convention. Any other file counts as a panel only when it actually builds a `Panel::make()` chain, so pointing `panel_paths` at a whole source tree does not turn every class into a panel.
+
 ## Graph Node Types
 
 | Node | Accent Color | Represents |
--- a/config/laravel-brain.php
+++ b/config/laravel-brain.php
@@ -301,6 +301,39 @@ return [
     ],
 
     // -------------------------------------------------------------------------
+    // Filament Search Paths
+    // -------------------------------------------------------------------------
+    // Directories (relative to the project root) scanned for Filament panels and
+    // for the resources, pages, widgets and relation managers they expose.
+    //
+    // An entry is used as-is when it is a directory and expanded as a glob pattern
+    // otherwise, so a modular monolith that keeps no app/ directory can point these
+    // at its packages:
+    //
+    //   'panel_paths' => ['app-modules/*/src/Filament'],
+    //   'paths'       => ['app-modules/*/src/Filament'],
+    //
+    // panel_paths  Where panels are declared. A file named *PanelProvider.php is
+    //              treated as a panel by convention (that is what Filament's own
+    //              installer writes); any other file counts only when it actually
+    //              builds a Panel::make() chain, so pointing this at a whole source
+    //              tree does not turn every class into a panel.
+    //
+    // paths        Roots of the Filament class tree. Resources are recognised by a
+    //              Resources/ path segment, pages by Pages/, widgets by Widgets/ and
+    //              relation managers by RelationManagers/ — the layout Filament's
+    //              generators produce.
+    //
+    'filament' => [
+        'panel_paths' => [
+            'app/Providers/Filament',
+        ],
+        'paths' => [
+            'app/Filament',
+        ],
+    ],
+
+    // -------------------------------------------------------------------------
     // Livewire Component Search Paths
     // -------------------------------------------------------------------------
     // Directories (relative to project root) that are searched when resolving
--- a/src/Analysis/FilamentAnalyzer.php
+++ b/src/Analysis/FilamentAnalyzer.php
@@ -94,11 +94,122 @@ class FilamentRelationManagerDefinition
 
 class FilamentAnalyzer
 {
+    /**
+     * Where panels are declared in a default Laravel skeleton.
+     *
+     * @var string[]
+     */
+    public const DEFAULT_PANEL_PATHS = ['app/Providers/Filament'];
+
+    /**
+     * Where resources, pages, widgets and relation managers live in a default
+     * Laravel skeleton.
+     *
+     * @var string[]
+     */
+    public const DEFAULT_PATHS = ['app/Filament'];
+
     private PhpFileParser $parser;
 
-    public function __construct()
-    {
+    /** @var string[] */
+    private array $panelPaths;
+
+    /** @var string[] */
+    private array $paths;
+
+    /**
+     * class FQCN => parent FQCN, collected from the scanned Filament trees so a
+     * resource or relation manager that extends a project base class is still
+     * recognised.
+     *
+     * @var array<string, string>
+     */
+    private array $classParents = [];
+
+    /**
+     * @param  string[]  $panelPaths  directories holding panel definitions, relative to the
+     *                                project root; glob patterns are expanded
+     * @param  string[]  $paths  directories holding resources, pages, widgets and relation
+     *                           managers, relative to the project root; glob patterns are
+     *                           expanded
+     */
+    public function __construct(
+        array $panelPaths = self::DEFAULT_PANEL_PATHS,
+        array $paths = self::DEFAULT_PATHS,
+    ) {
         $this->parser = new PhpFileParser;
+        $this->panelPaths = $panelPaths;
+        $this->paths = $paths;
+    }
+
+    /**
+     * Expand the configured relative paths into absolute directories that exist.
+     * An entry is used verbatim when it is a directory and treated as a glob
+     * pattern otherwise, so a pattern such as `app-modules/*` + `/src/Filament`
+     * works for a modular monolith that keeps no `app/` directory at all.
+     *
+     * @param  string[]  $relativePaths
+     * @return string[] absolute directories
+     */
+    private function resolveDirs(string $projectRoot, array $relativePaths): array
+    {
+        $dirs = [];
+
+        foreach ($relativePaths as $relativePath) {
+            $relativePath = trim((string) $relativePath);
+            if ($relativePath === '') {
+                continue;
+            }
+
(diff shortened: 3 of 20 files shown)


SanderMuller (changes requested):
The configurable paths, the panel guard and the cycle safety all check out. One thing does not, and it is
the headline feature, so I am holding rather than approving.

## A base class in the same namespace is still missed

`extends`-chain recognition works when the base class is imported. It does not when the base sits in the
same namespace as its children, which needs no `use` statement and is how most people write it:

```php
namespace App\Filament\Resources;

abstract class AppResource extends Resource {}          // found
class OrderResource extends AppResource {}              // NOT found — no use statement needed, none written
```

```
resources found: ["AppResource"]
```

Add a redundant `use App\Filament\Resources\AppResource;` to the child and both appear. The parent map
shows why:

```
App\Filament\Resources\AppResource     => Filament\Resources\Resource
App\Filament\Resources\OrderResource   => AppResource            <- bare short name
```

`collectClassParents()` stores `$this->useMap[$parent] ?? $parent`, so a parent with no import stays a
short name and can never match a key in a map that is keyed by FQCN. `descendsFrom()` then looks up
`classParents['AppResource']`, finds nothing, and returns false.

This is the failure mode the PR opens by describing: not wrong, quiet. A modular application whose module
base class lives beside its resources gets the same silent zero the PR is fixing everywhere else.

**It is in three places, not one** — the same trap you flagged so precisely in #90. `grep` for
`useMap[$parent] ?? $parent`: `collectClassParents()`, `extractResource()` and `extractRelationManager()`
each resolve the parent independently, and fixing only the map changes nothing, because the short name is
handed to `descendsFrom()` from the other two.

`PhpFileParser::resolvedName()` already covers this — it is what #73 used for the same class of bug. I
patched all three sites with

```php
ltrim(PhpFileParser::resolvedName($node->extends) ?? ($this->useMap[$parent] ?? $parent), '\\')
```

and the same-namespace case resolves, with your 23 Filament tests still passing. A fixture whose base class
sits beside its children would pin it.

## What is right

**The gate.** Byte-identical on two of three applications. The third gains exactly one widget node and one
edge, nothing lost: a widget under `app/Filament/<PanelName>/Widgets/`, which is neither of the two
locations the old hand-rolled passes looked in. That is the single-pass `/Widgets/` filter working, and it
is worth noting the compatibility section says no behaviour change for a default layout — this is one, in
the good direction.

**The cycle is safe.** I went looking for the `isAuthClass` problem from #51, where a cyclic `extends`
chain exhausted memory. `descendsFrom()` has a depth cap of 20, and two classes extending each other
inside a scanned tree complete in 15 ms rather than hanging.

**The panel guard holds.** A plain class inside a configured panel path is not treated as a panel, and a
panel built through an aliased import (`use Filament\Panel as FilamentPanel`) is still found. Both
behaviours you claimed, both confirmed.

`composer test`: 262 passed, PHPStan 0, Pint clean. CI has not run on this branch, so that is local against
`82efb6c9`.

## What would change my mind

The three-site resolution fix, or an argument that the same-namespace base class is not worth supporting.
Everything else here I would take as is.

webard:
You are right, and it was live in the codebase I developed this against — I just never looked, because I only ever compared the count against a number I had derived from the same broken assumption.

`app-modules/inventory/.../Suppliers/RelationManagers/` has three managers extending `AbstractInventoryMovementsRelationManager`, which sits in that same namespace with no import anywhere. Isolating the change on that tree:

```
without the fix:  relationManagers 22
with the fix:     relationManagers 25      (resources and widgets unchanged)
```

Three real relation managers were invisible and I shipped the PR saying the extends chain was handled.

**Fixed at all three sites**, as you said it had to be. `collectClassParents()`, `extractResource()` and `extractRelationManager()` now share one resolver built on `PhpFileParser::resolvedName()` with the use map as fallback. Patching fewer would have changed nothing, because the two extractors hand `parentFqcn` straight to `descendsFrom()`.

Two fixture classes pin it — a resource and a relation manager whose base class sits beside them, neither importing it. I checked both directions: they fail with the resolver reverted to the use map alone, and only those two fail.

`262 → 264 passed`, PHPStan 0, Pint clean. `6503371`.

Two things I want to note rather than let pass:

**This is the second time I have written the same class of bug** — in #89 I used `resolvedName()` for `Schedule::job(Job::class)` for exactly this reason, and then did not reach for it here. The tell both times is a name read from source being compared against a map keyed by FQCN.

**And the third time a test of mine asserted a true thing without pinning the mechanism.** `it recognises a resource that extends a project base class` passes because the fixture's base is imported. It never exercised the branch that matters. That is the same shape as the vendor-exclusion fixture in #89, and it is worth me checking for directly: does the fixture distinguish the two implementations, or merely satisfy the one I wrote?

On the widget node your gate found — a widget under `app/Filament/<PanelName>/Widgets/` that the old two hand-rolled passes missed — agreed that the compatibility section should not have said "no behaviour change for a default layout". That is a second behaviour change in the same PR, in the good direction, and it should be stated rather than discovered.

SanderMuller (approved):
Re-reviewed at `6503371d`. I re-ran the probe that produced the finding rather than reading the fix for it: your new fixture, driven through the analyzer, against both the head I reviewed and this one.

```
base (82efb6c9)   resources 1  [CatalogResource]
                  relationManagers 1  [CatalogRelationManager]

head (6503371d)   resources 2  [CatalogResource, ProductResource]
                  relationManagers 2  [CatalogRelationManager, PricesRelationManager]
```

Only the abstract bases were found before, which is the shape I described: the child's parent is a bare short name and never matches a key in a map of FQCNs. Both children are found now, at both sites.

Putting it in one static rather than patching three call sites is the better fix. I checked there is no fourth: `parentFqcn()` is used at all three places that resolve an `extends` clause to an FQCN, and the two remaining reads of `->extends` take `getLast()` deliberately, to `match` a direct base short name onto a page or widget type label. Those pick a label rather than walk a chain, so the short name is the right thing there.

**The tests bite, and only where they should.** Reverting `parentFqcn()` to the use-map-only form fails exactly the two new tests, with the other 22 in that file still passing, so nothing about the imported-base path moved. Mutation confirmed with `git diff --stat` before I read the run.

**The gate cannot see this fix, and that is worth saying rather than reporting as a pass.** Re-run on three applications: byte-identical on two, and the third gains the same one widget node and one edge I characterised before (`nodeCount` 2874 to 2875, `edgeCount` 5196 to 5197, a widget under a panel-named directory, nothing lost). No new differences. But none of the three contains a resource or relation manager whose base sits unimported in its own namespace, which I checked directly rather than inferring from the silence, so the gate has nothing to say about the change this PR is actually making. The fixture is the evidence here, not the corpora.

One thing the probe turned up that is *not* a problem: application C has a Filament page extending a base in its own namespace with no import. Pages are recognised by directory rather than by an extends chain, and that page appears identically in both dumps, so it was never affected.

`264 passed (892 assertions)`, PHPStan clean, Pint clean. Approving.

mrmarchone:
Thanks @webard for your support , could you please make all your PRs to main not v2.x branch ?

webard:
@mrmarchone Done — all sixteen are on `main` now: #88, #89, #90, #91, #98, #99, #100, #101, #102, #103, #104, #105, #106, #107, #108, #109.

Nothing had to be rebased, and it is worth saying why rather than just asserting nothing moved: `main` and `v2.x` are the same commit right now (`344eeca`), so the merge base is unchanged and every diff is byte-for-byte what it was. Checked after retargeting — commit counts identical on all sixteen, and all sixteen still report `MERGEABLE`. No PR description referenced `v2.x`, so nothing is stale.

One thing you may want to look at: the repo's default branch is still `v2.x` (`upstream/HEAD -> v2.x`), which is why new PRs land there unless the base is set by hand. Switching the default in the repo settings would make `main` the target automatically and save you asking again.

Four of these are stacked, so merge order matters: #99 → #100 → #101 → #107 → #108 → #109 are one chain, and #104 sits on #103. Each says so at the top of its description. If you would rather have them as independent PRs against `main`, say the word — they are frontend-only and I can flatten them, it just means the later ones carry the earlier diffs until the earlier ones land.

mrmarchone (approved):

PR #89, opened 2026-08-27
Feat: modular psr4 and command attributes

webard:
## Resolve classes in package-based applications, and recognise commands, schedules and channels however they are declared

Four independent places assume the default Laravel skeleton and go quiet - not wrong, quiet - the moment an application does not match it. A scan still succeeds and still prints a summary; the summary just says `0`.

### 1. The PSR-4 map is read from the root composer.json only

`ControllerAnalyzer::buildPsr4Map()` and `ModelAnalyzer::buildPsr4Map()` each read `autoload`/`autoload-dev` out of the project's own `composer.json`. In an application whose code lives in packages, that file declares `App\ => app/` and nothing else, so no class in any package can be mapped to a file.

The failure mode is silent and total. `ModelAnalyzer::discoverModels()` walks the configured directories and finds every model; `analyze()` then drops all of them, because `resolveFile()` has no namespace to match. On a ~90-package monolith that is 174 models discovered and 0 kept - and with them the call chains, jobs, events, listeners and DB queries that hang off models. Nothing in the output suggests a lookup problem.

The first of those two maps is the canonical one: `ProjectAnalyzer` builds it once and hands it to `MethodTracer`, `ListenerAnalyzer`, `PolicyAnalyzer`, `QueryTracer` and the Filament page tracer. So this is one fix, not six.

A new `ComposerPsr4Map` merges four sources and both old builders delegate to it:

- the root `composer.json`, exactly as before;
- packages Composer installed from a `path` repository, read from `vendor/composer/installed.json` - the authoritative list of what is actually wired up;
- `path` repositories declared in the root `composer.json`, read from each package's own `composer.json`, which covers a checkout where `composer install` has not run and a package that is present but not required;
- the nwidart/laravel-modules `Modules/` scan, moved over unchanged from `ControllerAnalyzer`.

**Regular vendor dependencies are deliberately left out.** Reading `vendor/composer/autoload_psr4.php` would have been one line and a complete map, but making vendor classes resolvable lets `MethodTracer` walk into framework and library internals, which is not what the graph is for. Local packages are enough.

### 2. A command is only recognised by its `$signature` property

`ConsoleAnalyzer` reads `protected $signature = '...'` and nothing else. A command that declares itself with Laravel 12's `#[Signature]` / `#[Description]` attributes, or with Symfony's `#[AsCommand]`, has no such property and is skipped entirely - 109 of 111 commands in the application I measured on.

All three forms are now read, along with the pre-signature `protected $name`. A property wins over an attribute, matching Laravel's own precedence.

### 3. Scheduled tasks are only looked for in files named `*console*`

`console_route_paths` is documented as a glob, but the analyzer truncates the pattern at its first wildcard and then keeps only files whose basename contains `console`. Laravel's own skeleton keeps the schedule in `routes/console.php`, so this works - until a schedule is split into `routes/schedule.php`, which nothing ever reaches. That was 40 files and 61 entries in the application I measured on.

`schedule` is now accepted alongside `console`. Two things follow from finally reading those files:

- `Schedule::job()` and `Schedule::call()` are recorded next to `Schedule::command()`; a job target is resolved to its FQCN through the parser's resolved-name attribute rather than the short name it was written with.
- `walkChainForFrequency()` was a stub returning `''`, so every statically-registered entry carried an empty cadence. Frequency lives on the calls wrapped *around* the registration and a node cannot see its own parents, so the chain is now read on the way in - traversal is top-down - and parked for the static call that arrives a few nodes later. `Schedule::command('x')->dailyAt('05:30')->withoutOverlapping()` now reports `dailyAt`.

### 4. Channels are only recognised as `Broadcast::channel()`

An application that wraps the facade - to scope every channel to a tenant, for instance - registers all of its channels through its own class and none through `Broadcast`, so none are found.

A `channel_registrars` config key names additional classes whose static `channel()` call registers a channel. `Broadcast` is always recognised, so the default is unchanged. Entries are matched by short class name, so both `'TenantChannel'` and `App\Broadcasting\TenantChannel::class` work.

### Tests

17 new tests across `ComposerPsr4MapTest`, `ConsoleAnalyzerTest` and `ChannelAnalyzerTest` - the latter two did not exist - against two new fixtures:

- `modular-project`: a package installed from a `path` repository, a package declared as a `path` repository but never installed, a regular vendor dependency as a negative control, four spellings of a command declaration (`#[Signature]`+`#[Description]`, `#[AsCommand]`, legacy `$name`, and one declaring both a property and an attribute to pin the precedence), a `routes/schedule.php` and a `routes/channels.php` registering through both the facade and a wrapper;
- `nwidart-project`: pins the conventional `Modules\{Name}` mapping that moved out of `ControllerAnalyzer`.

Full suite: `271 passed (883 assertions)`, up from 254 with no existing test changed. `pint` clean, `phpstan` 0 errors.

A ~90-package monolith with no `app/` directory, same scan before and after:

| | before | after |
|---|---|---|
| controllers | 0 | 223 |
| lifecycle call edges | 145 | 1370 |
| models | 0 | 177 |
| commands | 2 | 109 |
| scheduled tasks | 0 | 58 |
| broadcast channels | 0 | 6 |
| command call chains | 0 | 768 |
| DB query actions | 0 | 87 |
| **graph** | **1339 nodes / 5668 edges** | **2865 nodes / 8491 edges** |

Before, the graph held routes, middleware and controllers and nothing else. After, it holds 593 service, 327 model, 113 command, 54 schedule, 38 listener, 34 validation-request, 30 event, 28 job and 4 channel nodes. Total scan time went from 17.5 s to 21.5 s.

### Compatibility

The PSR-4 map is a superset of what it was; `channel_registrars` defaults to empty and `Broadcast` stays built in; command attributes are consulted only when no property declares the same thing; and `console` remains an accepted filename keyword.

**But a default Laravel layout is not untouched, and this is the part to read before merging.** Declaring a command with Laravel 12's `#[Signature]` is a supported default; such a command was previously invisible to the scan, and reading it changes what the command *is*:

```
- "id": "cmd-acme-console-commands-backupdatabase"   label: Acme\Console\Commands\BackupDatabase
+ "id": "cmd-acme-backup-database"                   label: acme:backup-database
```

The label becomes the command a person actually types, which is the point. But the tab id is derived from it, so **every command that gains a signature is re-keyed, along with the call-chain hops beneath it** — measured on one application with five such commands: 173 stored subgraph blobs became unaddressable and 178 new ones took their place.

Those ids are what the viewer lists and what a saved link points at, so a bookmark to one of those tabs stops resolving. The orphaned blobs are cleaned up by #98 if that lands first; without it they stay on disk.

An application with no attribute-declared commands sees none of this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -251,6 +251,29 @@ php artisan brain:scan --auto-discover
 
 > **Heads up:** in auto-discover mode the source file and line number of each route are not available, so the sidebar will not group routes by their declaring file (everything falls under a single group). Use the default AST mode if file/line grouping matters to you.
 
+### Modular applications
+
+Class names are resolved through the PSR-4 map built from the root `composer.json` **and** from every local package: packages installed from a `path` repository, and nwidart/laravel-modules under `Modules/`. An application whose code lives in packages rather than in `app/` therefore resolves normally — without it, classes are discovered and then dropped, because no file can be found for them.
+
+Regular vendor dependencies are deliberately left out of that map, so the call-chain tracer never walks into framework or library internals.
+
+### Artisan commands
+
+A command is recognised however it names itself: a `$signature` (or the older `$name`) property, Laravel's `#[Signature]` / `#[Description]` attributes, or Symfony's `#[AsCommand]`. A property wins over an attribute, matching Laravel's own precedence.
+
+Scheduled tasks are read from `routes/console.php` and from a schedule split into its own `routes/schedule.php`, in both the `Schedule::command()` and `$schedule->command()` spellings, along with `job()` and `call()` entries and the cadence each one is chained with.
+
+### Broadcast channels
+
+Channels registered through the `Broadcast` facade are found out of the box. If your application registers them through a wrapper of its own — one that scopes every channel to a tenant, for instance — name that class in `channel_registrars` and its `::channel()` calls are read the same way:
+
+```php
+// config/laravel-brain.php
+'channel_registrars' => [
+    App\Broadcasting\TenantChannel::class,
+],
+```
+
 ### Call chain tracing
 
 From each controller action (and Filament page method), the tracer follows:
--- a/config/laravel-brain.php
+++ b/config/laravel-brain.php
@@ -180,6 +180,23 @@ return [
     ],
 
     // -------------------------------------------------------------------------
+    // Channel Registrars
+    // -------------------------------------------------------------------------
+    // Classes whose static channel() call registers a broadcast channel, on top of
+    // the Broadcast facade, which is always recognised.
+    //
+    // Add your own wrapper here if channels are not registered through the facade
+    // directly — a class that scopes every channel to a tenant, for example. Entries
+    // are matched by short class name, so either spelling works:
+    //
+    //   'channel_registrars' => [
+    //       'TenantChannel',
+    //       App\Broadcasting\TenantChannel::class,
+    //   ],
+    //
+    'channel_registrars' => [],
+
+    // -------------------------------------------------------------------------
     // Job Dispatch
     // -------------------------------------------------------------------------
     // Brain resolves the built-in dispatch verbs (dispatch(), dispatch_sync(),
--- a/src/Analysis/ChannelAnalyzer.php
+++ b/src/Analysis/ChannelAnalyzer.php
@@ -25,14 +25,35 @@ class ChannelAnalyzer
     /** @var string[] */
     private array $channelPaths;
 
+    /** @var string[] class names, short or fully qualified */
+    private array $registrars;
+
     /**
      * @param  string[]  $channelPaths  Glob patterns relative to the project root.
      *                                  Only files whose basename contains "channel" are parsed.
+     * @param  string[]  $registrars  Additional classes whose static channel() call registers
+     *                                a channel, on top of Broadcast. An application that wraps
+     *                                the facade — to scope channels to a tenant, say — registers
+     *                                every one of its channels through its own class, and none
+     *                                of them are Broadcast::channel() calls.
      */
-    public function __construct(array $channelPaths = ['routes/*/*.php'])
+    public function __construct(array $channelPaths = ['routes/*/*.php'], array $registrars = [])
     {
         $this->parser = new PhpFileParser;
         $this->channelPaths = $channelPaths ?: ['routes/*/*.php'];
+        $this->registrars = array_values(array_unique(array_merge(
+            ['Broadcast'],
+            array_map(static fn (string $registrar): string => self::shortName($registrar), $registrars),
+        )));
+    }
+
+    /** Compare registrars by short name: an import may be aliased, a call may be qualified. */
+    private static function shortName(string $class): string
+    {
+        $class = trim($class, '\\');
+        $position = strrpos($class, '\\');
+
+        return $position === false ? $class : substr($class, $position + 1);
     }
 
     /**
@@ -100,13 +121,17 @@ class ChannelAnalyzer
     private function extractChannels(array $ast, array $useMap, string $file): array
     {
         $traverser = new NodeTraverser;
-        $visitor = new class($file, $useMap) extends NodeVisitorAbstract
+        $visitor = new class($file, $useMap, $this->registrars) extends NodeVisitorAbstract
         {
             public array $channels = [];
 
+            /**
+             * @param  string[]  $registrars
+             */
             public function __construct(
                 private string $file,
                 private array $useMap,
+                private array $registrars,
             ) {}
 
             public function enterNode(Node $node): ?int
@@ -119,7 +144,7 @@ class ChannelAnalyzer
                 if (! $node->class instanceof Node\Name) {
                     return null;
                 }
-                if ($node->class->getLast() !== 'Broadcast') {
+                if (! in_array($node->class->getLast(), $this->registrars, true)) {
                     return null;
                 }
 
--- a/src/Analysis/ComposerPsr4Map.php
+++ b/src/Analysis/ComposerPsr4Map.php
@@ -1,0 +1,216 @@
+<?php
+
+declare(strict_types=1);
+
+namespace LaraMint\LaravelBrain\Analysis;
+
+/**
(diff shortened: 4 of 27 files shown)


SanderMuller (approved):
Gated on three production applications and classified everything that moved. Approving, with two things
worth reading before merge.

## What the output gate showed

| application | result |
|---|---|
| A | 173 ids removed, 178 added |
| B | nothing lost, 73 added |
| C | byte-identical |

B is the clean case: it has one `path` repository, its package now resolves, and six commands plus their
call chains appear. Pure addition.

A is the one to look at. It has no `path` repositories at all and is the most default-shaped of the three,
so by the compatibility section it should not have moved. It moved because five commands declare
themselves with attributes rather than `$signature`, and reading them changes what those commands *are*:

```
- "id": "cmd-acme-console-commands-backupdatabase"   label: Acme\Console\Commands\BackupDatabase
+ "id": "cmd-acme-backup-database"                   label: acme:backup-database
```

(names changed, shape exact)

That is better. The label goes from a class name to the command anyone actually types. But the id is
derived from it, so five commands are re-keyed and roughly 168 call-chain hops underneath them are re-keyed
with them, and 173 stored subgraph blobs become unaddressable.

## So the compatibility claim is too strong

"No behaviour change for a default Laravel layout" does not hold for an application using Laravel 12
command attributes, which is a default-supported way to declare a command. The change is right; the
sentence promises more than it can. Worth rewording, and worth saying explicitly that tab ids change for
any command that gains a signature, because those ids are what the viewer lists and what a link points at.
#98 prunes the orphaned blobs, if that lands first.

## One test does not pin the decision it names

`leaves regular vendor dependencies out of the map` still passes with the `dist.type === 'path'` filter
removed entirely. The fixture's `acme/vendor-lib` has no directory under `vendor/acme/`, so
`localPackageRoot()` returns null whatever the filter does, and the deliberate exclusion you call out in
the description is unguarded.

Verified fix, two files:

```
tests/fixtures/modular-project/vendor/acme/vendor-lib/composer.json
{ "name": "acme/vendor-lib", "autoload": { "psr-4": { "Acme\\VendorLib\\": "src/" } } }
```

With that present the test passes as written and fails when the filter is removed. I checked both
directions.

The precedence test is the opposite case and holds up: reversing `$this->signature ?? $this->attributeSignature`
kills `lets a $signature property win over a Signature attribute`.

## The cost

Interleaved against the merge-base, three reps, with a null arm establishing the noise floor:

| application | before | after | noise floor |
|---|---:|---:|---:|
| A | 2,141 ms | 2,152 ms | 124 ms |
| B | 1,399 ms | 1,404 ms | 47 ms |

Both deltas are inside the floor, so no measurable cost here. That is consistent with your 17.5 s to
21.5 s rather than contradicting it: the cost tracks how much more the scan finds, and on these
applications it finds almost nothing new. An application where the map goes from empty to 90 packages pays
for the graph it gains.

## Limits of this evidence

None of these three is a package-based application — one `path` repository between them, no `Modules/`,
no `routes/schedule.php`. So the largest claims in the description, the ones that take controllers from 0
to 223, are ones I cannot exercise. What I can say is that the change does not damage a conventional
layout, and that where a `path` package does exist it is picked up correctly.

`composer test` on this head: 271 passed, PHPStan 0, Pint clean. CI has not run on this branch, so that
green is mine, local against `3f55a1d5`.

webard:
Both taken, and the second one is the more useful finding.

**The test did not pin what it claimed.** You are right that `acme/vendor-lib` having no directory under `vendor/acme/` meant `localPackageRoot()` returned null regardless of the filter, so the exclusion was unguarded. The fixture package exists on disk now, and I checked it the way you did: the test passes as written and fails with the `dist.type === 'path'` filter removed. `439ef71`.

That is the second time in this series a test of mine asserted a true thing without pinning the mechanism, and both times the tell was the same — the fixture could not distinguish the two branches. Worth me checking for directly rather than waiting to be told.

**The compatibility claim was too strong** and now says so. `#[Signature]` is a supported default, a command declared that way was invisible before, and reading it re-keys the tab: your 173 removed / 178 added is the concrete form of that, and I have quoted it. The section now says plainly that the ids change, that a saved link to one of those tabs stops resolving, and that #98 cleans up the orphaned blobs if it lands first.

**On the cost measurement:** a null arm to establish the noise floor, and both deltas inside it — that answers a question I had left open, since my own before/after was 17.5 s to 21.5 s on an application where the scan finds a great deal more. Consistent rather than contradictory, exactly as you put it.

And noted on the limits: one `path` repository between three applications means the claims that take controllers from 0 to 223 are not ones your corpora can exercise. That the change does not damage a conventional layout is the more valuable half of the result anyway.

SanderMuller (approved):
Re-reviewed at `439ef719`. My earlier approval described `3f55a1d5`, so this replaces it.

The delta is two fixture files and nothing else, which I checked rather than assumed:

```
tests/fixtures/modular-project/vendor/acme/vendor-lib/composer.json
tests/fixtures/modular-project/vendor/acme/vendor-lib/src/Client.php
```

`src/` is untouched, so the gate classification and the timings in my earlier review still describe this head.

**The test pins its decision now.** I re-ran the mutation that produced the finding, on both heads:

```
replace  ($package['dist']['type'] ?? null) !== 'path'   with   nothing

3f55a1d5   6 passed          <- the filter could be deleted and the suite stayed green
439ef719   1 failed, 5 passed <- "leaves regular vendor dependencies out of the map"
```

`installed.json` already named `acme/vendor-lib`, but with no directory behind it the builder skipped the package for a reason that had nothing to do with the filter. Giving it a real `composer.json` and a class makes the filter the only thing keeping it out, which is what the test's name claims. Nothing else in that file moved under the mutation.

**The compatibility wording.** The description now says a default layout is not untouched, names Laravel 12's `#[Signature]` as the case, and states plainly that every command gaining a signature is re-keyed along with the call-chain hops beneath it, with the 173-for-178 count. That is the thing a maintainer needs to read before merging, and it is now in the place they will read it.

`271 passed (883 assertions)`, PHPStan clean, Pint clean. Branch is level with `v2.x`. Approving the current head.

mrmarchone (approved):

PR #90, opened 2026-08-27
Feat: read service providers and facades from configurable paths

webard:
## Read service providers and facades from wherever they live

`ContainerBindingAnalyzer` looks in `app/Providers` and `FacadeAnalyzer` in `app/`, both hardcoded. An application that keeps its providers and facades in packages therefore reports `0 bindings` and `0 facades` - not an error, just an empty result - and with them the graph loses two kinds of edge: the one that wires an interface or abstract class through to the concrete class bound to it, and the one that wires a facade through to the class behind its accessor.

Both paths become configurable, defaulting to what they are today.

```php
// config/laravel-brain.php
'container_bindings' => [
    'provider_paths' => ['app-modules/*/src'],
],

'facades' => [
    'paths' => ['app-modules/*/src'],
],
```

An entry is used as-is when it names a directory and expanded as a glob pattern otherwise. A shared `SourceDirectories` helper does that expansion and the recursive PHP-file walk for both analyzers, de-duplicating by real path so overlapping patterns cannot scan a file twice.

### Two details that are easy to miss

**`FacadeAnalyzer` hardcodes `app/` in two places, not one.** Besides the directory it iterates, the by-short-name lookup it uses to follow a facade's parent chain is pinned to `['app']`. Fixing only the first finds facades that extend Illuminate's `Facade` directly and loses every facade that reaches it through an application-level base class - exactly the case the analyzer's second pass exists to handle. Both now follow the configured directories, and `findFileInAppDir()` is renamed to `findFileInSourceDirs()` to stop the old assumption reading as intentional. There is a test for a facade whose base class lives in a *different* package.

**`ContainerBindingAnalyzer` never reached deeply nested providers, even under the default path.** The scan was `app/Providers/*.php` plus `app/Providers/**/*.php`, and PHP's `glob()` does not cross directory separators - `**` matches a single segment like `*` does. So `app/Providers/Domain/Billing/BillingServiceProvider.php` was silently skipped. The walk is now recursive, which fixes that independently of the path change; a fixture provider sits two levels down to pin it.

### Tests

9 new tests, no existing test changed, against a new `packaged-app` fixture: two packages, a provider nested two levels deep, a facade with a `::class` accessor, a facade whose base class lives in another package, and an abstract base as a negative control. Each analyzer gets a test asserting that the *same* fixture yields nothing under the default paths, and one asserting that `app/` still works when it is among the configured paths.

Full suite: `263 passed (870 assertions)`, up from 254. `pint` clean, `phpstan` 0 errors.

A ~90-package monolith with no `app/` directory:

| | before | after |
|---|---|---|
| container bindings | 0 | 57 |
| facades | 0 | 4 |

Node and edge totals are unchanged in that run, which is what should happen: these registries do not create nodes, they wire existing ones. `maybeWireContainerBinding()` fires only when a call-chain edge lands on an interface or abstract class, and in the application measured the call chains are still thin for an unrelated reason (its classes live in packages the PSR-4 map does not cover - a separate change). The 57 bindings are real and correctly attributed; what they wire up grows as the call graph does.

### Compatibility

Defaults are the previous hardcoded paths, so a standard Laravel application scans exactly as before, with one deliberate exception: a provider nested more than one directory below `app/Providers` is now found.

That is a behaviour change on a default installation, not only on the layouts this PR is about. An application that keeps providers in, say, `app/Providers/Domain/Billing/` gains the bindings they register and the interface-to-concrete edges that follow from them — a graph that grew after upgrading, with nothing in the configuration changed, is explained by this. It is the behaviour the glob was reaching for and never achieved, so the fix is right; it is worth knowing it is visible.

Both analyzer constructors gain one optional argument after the existing `?PhpFileParser`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -284,6 +284,23 @@ From each controller action (and Filament page method), the tracer follows:
 
 This produces the full edge list used to build the graph.
 
+### Service providers and facades
+
+Container registrations (`bind()`, `singleton()`, `scoped()` and their `*If` variants, plus the `$bindings` property) are read from service providers, and application-level facades — classes whose inheritance chain reaches `Illuminate\Support\Facades\Facade` — from the application's source tree. Both feed the graph: an edge that lands on an interface or an abstract class is wired through to the concrete class bound to it, and a facade call is wired through to the class behind its accessor.
+
+Both directories default to the standard skeleton and take glob patterns, so an application whose providers and facades live in packages points them at its own layout:
+
+```php
+// config/laravel-brain.php
+'container_bindings' => [
+    'provider_paths' => ['app-modules/*/src'],
+],
+
+'facades' => [
+    'paths' => ['app-modules/*/src'],
+],
+```
+
 ### Filament PHP support
 
 When Filament is installed, the scanner discovers every panel registered via service providers, then resolves its resources, pages, widgets, and relation managers — both explicitly listed (`->resources([...])`) and auto-discovered (`->discoverResources(for: '...')`). Filament page methods are traced through the same call-chain engine as controller actions, so models and services they touch appear in the graph.
--- a/config/laravel-brain.php
+++ b/config/laravel-brain.php
@@ -351,6 +351,42 @@ return [
     ],
 
     // -------------------------------------------------------------------------
+    // Container Binding Search Paths
+    // -------------------------------------------------------------------------
+    // Directories holding service providers, scanned recursively for container
+    // registrations: bind() / singleton() / scoped() (and their *If variants) plus
+    // the $bindings property.
+    //
+    // An entry is used as-is when it is a directory and expanded as a glob pattern
+    // otherwise, so an application whose providers live in packages can say:
+    //
+    //   'provider_paths' => ['app-modules/*/src'],
+    //
+    'container_bindings' => [
+        'provider_paths' => [
+            'app/Providers',
+        ],
+    ],
+
+    // -------------------------------------------------------------------------
+    // Facade Search Paths
+    // -------------------------------------------------------------------------
+    // Directories scanned for application-level facades — classes whose inheritance
+    // chain reaches Illuminate\Support\Facades\Facade. The same directories back the
+    // by-short-name lookup used to follow a facade's parent chain, so a base facade
+    // must be inside them too.
+    //
+    // Glob patterns are expanded, as above:
+    //
+    //   'paths' => ['app-modules/*/src'],
+    //
+    'facades' => [
+        'paths' => [
+            'app',
+        ],
+    ],
+
+    // -------------------------------------------------------------------------
     // Livewire Component Search Paths
     // -------------------------------------------------------------------------
     // Directories (relative to project root) that are searched when resolving
--- a/src/Analysis/ContainerBindingAnalyzer.php
+++ b/src/Analysis/ContainerBindingAnalyzer.php
@@ -18,35 +18,48 @@ use PhpParser\NodeTraverser;
 use PhpParser\NodeVisitorAbstract;
 
 /**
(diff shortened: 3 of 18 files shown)


SanderMuller (approved):
Reviewed on its own merits, which is overdue: this commit is also carried by #91, and approving that one
already approved this. Reviewing it after the fact is the wrong order, and it is mine to have got wrong.
It holds up.

## The nested-provider bug is real, and I could only prove it by constructing it

The output gate is byte-identical on three production applications, and that proves less here than it
usually does. All three keep providers more than one directory below `app/Providers` — 7, 1 and 2 of them —
so the fix should bite, and the graph does not move because those particular files are Filament panel
providers rather than `bind()` calls. The registries are also unchanged: 7, 20 and 16 bindings, and 0, 6
and 0 facades, before and after.

A null result on my corpora is evidence about my corpora, so I built the case instead: one provider at the
top level and one two directories down, each binding an interface.

```
before this PR:  1 binding   ["TopInterface"]
with this PR:    2 bindings  ["LedgerInterface", "TopInterface"]
```

Your reading of `glob()` is right, and the fix is independent of the path change, exactly as you say.

## The second `app/` hardcode

I checked this the way you framed it, by reverting only the parent-chain lookup to `['app']` and leaving
the directory iteration configurable. `follows a facade parent chain that crosses package boundaries`
fails. So a half-fix really does lose facades that reach the base through an application-level class, and
the test pins it.

`reaches a provider nested more than one directory deep` dies too when the walk goes back to
non-recursive. Both bite.

## Also checked

`SourceDirectories::resolve()` is called once per `analyze()` in each analyzer and kept in a field, not
per class lookup. The memoisation finding from #91 does not apply here — worth saying, since the same
helper is involved and it would be reasonable to assume it carries over.

The de-duplication claim holds: `packages/*/src` together with `packages/a/src` resolves to one directory
and yields each file once.

`composer test` on this head: 263 passed, PHPStan 0, Pint clean.

## Gates

Output gate applied because this touches `src/Analysis`, and it is byte-identical on all three
applications — with the caveat above about what that does and does not establish. No benchmark, since
nothing here runs per class lookup. CI has not run on this branch, so the green is mine, local against
`1e9c3fd5`.

## One thing worth deciding deliberately

Compatibility is described as preserved except that nested providers are now found. That is a real
behaviour change on a default installation, and the right one, but it means an application with a provider
two levels down gets bindings it did not have before and edges that follow from them. It changed nothing
on the three applications here, so I cannot say how visible it is in general — only that it is intended,
and that anyone surprised by new binding edges after upgrading should be pointed at this.

webard:
Thanks — and the point about your corpora being a null result is the part I want to pick up.

**On the deliberate decision:** you are right that it deserves more than a subordinate clause, and the compatibility section now says it plainly. A default installation with providers in, say, `app/Providers/Domain/Billing/` gains bindings and the interface-to-concrete edges that follow, with nothing in the configuration changed. Someone whose graph grew after upgrading should find the explanation there rather than have to bisect for it.

**Constructing the case rather than reporting a null result** is the right move, and I should have done it myself. My evidence for the glob defect was a reading of `**` plus a fixture provider two levels down; yours is 1 binding → 2 on a case built to discriminate. That is the stronger form and I will reach for it first next time.

**The half-fix check is the one I could not have run** — reverting only the parent-chain lookup and watching `follows a facade parent chain that crosses package boundaries` fail. That was the claim I was least able to demonstrate when I wrote it, and it is now pinned by exactly the test I hoped would pin it.

**Noted on memoisation.** `SourceDirectories::resolve()` is called once per `analyze()` here and held in a field, so #91's finding does not carry over — worth you saying so, because the shared helper does invite the assumption.

mrmarchone (approved):

mrmarchone:
@webard there is a small conflict here.

PR #91, opened 2026-08-27
Feat: modular view watch and lookup paths

webard:
## Configurable Blade view roots, watch paths and class-lookup paths

> Builds on #90, which introduces the `SourceDirectories` helper this reuses. Merge that one first.

Three more places assume the default Laravel skeleton. Each is a different consequence of the same assumption, so they are grouped rather than split.

### 1. Blade views are read from `resources/views` only

`BladeViewAnalyzer` takes a views directory as a constructor argument, but `ProjectAnalyzer` never passes one and no config key backs it, so the parameter only ever serves tests - the directory is `resources/views` in practice. An application that keeps templates in packages gets no view-composition edges at all.

`views.paths` makes it a list, with glob expansion. Two things follow from that:

- An `@include` is linked when the template exists under **any** configured root, not only the one the including file came from - one package rendering another's partial is a real edge.
- `GraphBuilder::resolveBladePath()` resolves view names independently, with its own hardcoded set of candidates. The analyzer's own docblock says the two must stay in lockstep, and changing one alone breaks that: the analyzer would report a link to a template the builder cannot open. `GraphBuilder::setViewPaths()` now takes the same list. For a namespaced view the hint is bound to a directory at runtime, by whichever provider registered it, so it cannot be mapped deterministically; after the existing `Modules/` and `vendor/` candidates the roots are searched by file name, preferring a root whose own path carries the hint.

### 2. Watch mode and incremental rebuilds only look at `app/`

Four places agree that a project is `app`, `routes` and `config`: `ScanCommand::collectMtimes()`, the scoped-rescan guard in `ScanCommand::scopeForRescan()`, `BuildFingerprint::capture()` and `IncrementalAnalyzer`. In a package-based application `--watch` notices nothing, and the fingerprint hashes whichever of those three directories happens to exist.

Two keys, because these are two different questions:

```php
'watch_paths'  => ['app', 'routes', 'config'], // what can change the graph
'source_paths' => ['app', 'src'],              // where application classes live
```

A change confined to `source_paths` can take the scoped path; a change anywhere else in `watch_paths` - a route file, a config file - forces a full rebuild. `BuildFingerprint::capture()` expands globs too, so `watch_paths` accepts patterns.

While wiring this up: `IncrementalAnalyzer::touchesNonApp()` decided containment with `str_contains($f, '/app/')` - the loose test that `ScanCommand` warns against in a comment four files away, which calls every file "inside app/" when the project itself lives under a directory of that name. Both now go through one anchored `SourceDirectories::contains()`.

### 3. The last-resort class lookup searches `['app', 'src']`

When Composer's PSR-4 map cannot place a class name, Brain falls back to searching by file name, and to joining a prefix with the namespace path. Seven sites hardcode where: `MethodTracer`, `GraphBuilder`, `ControllerAnalyzer`, `QueryTracer`, `ListenerAnalyzer`, `PolicyAnalyzer` and `SecurityAnalyzer` (the last three spelling it as the `App\ => app/` convention).

All seven read `source_paths` now. The prefix list moves into `SourceDirectories::classFilePrefixes()`, which keeps `app/Http/Controllers/` while `app` is among the paths - dropping it would quietly change behaviour for a standard Laravel application.

### Tests

14 new tests, no existing test changed:

- `SourceDirectoriesTest` - expansion, de-duplication, prefix building, and containment anchored at the project root (with the case of a project living under a directory called `app`).
- `BladeViewAnalyzerTest` - glob expansion across roots, an include resolving under a different root, an include with no matching file anywhere, and the same fixture yielding nothing under the default path.
- `IncrementalAnalyzerTest` - a change inside the configured source paths takes the scoped path; a change in a watched-but-not-source directory forces a full rebuild.

Full suite: `277 passed (889 assertions)`, up from 263. `pint` clean, `phpstan` 0 errors.

A ~90-package monolith with no `app/` directory. With `source_paths` pointed at its packages, the last-resort lookup starts placing classes that were previously unresolvable:

| | before | after |
|---|---|---|
| graph nodes | 1339 | 2118 |
| graph edges | 5668 | 7314 |

View composition stayed at `0` there, and that turned out to be correct rather than a remaining gap: that application's package templates contain no `@include`/`@extends`/`@component` at all, and every `<x-...>` in them is a namespaced vendor component, which the analyzer skips by design. The 40 view roots do resolve and the 80 templates under them are scanned - there is simply nothing first-party to link. The hardcoded path was real, and this fixes it for anyone whose packages do compose their own views; it is not what produced that particular zero.

### Compatibility

Every default is the previous hardcoded value, so a standard Laravel application behaves as before - with one deliberate exception: incremental rebuilds now test containment anchored at the project root instead of by substring, which is stricter and matches what the neighbouring code already documented. `BladeViewAnalyzer::__construct()` changes from `string` to `string[]`; `ControllerAnalyzer`, `QueryTracer`, `ListenerAnalyzer`, `PolicyAnalyzer`, `SecurityAnalyzer`, `MethodTracer` and `IncrementalAnalyzer` each gain one optional argument.


SanderMuller:
Reviewed and measured against three production Laravel applications. The description is accurate and the compatibility claim holds — I could not make a standard application behave differently. Two findings, one of which I would fix before merging.

## What I verified

- `277 passed (889 assertions)`, PHPStan 0 errors, Pint clean, exactly as stated.
- **Full build output is byte-identical to `main` on all three applications** — graph, every subgraph, manifest, totals, tracer edge list. For a 989-line change touching seven class-lookup sites, that is the claim that matters and it survives.
- Also byte-identical **through a symlinked project root**, which I checked specifically because `SourceDirectories::phpFiles()` yields `getRealPath()` while `contains()` compares against an unresolved root. Those two never meet — `phpFiles()` only feeds scanning, `contains()` only containment — so the mismatch is harmless today. Worth a note in `phpFiles()` so nobody later routes a containment test through it.
- **No performance regression on the defaults**: medians over three interleaved runs are 1220→1212, 1822→1829, 1118→1126 ms, parse counts identical.

## 1. The glob resolution is unmemoised, and it sits in a per-FQCN loop

`SourceDirectories::classFilePrefixes()` and `resolve()` are called from inside `MethodTracer::resolveFile()`, and likewise from `GraphBuilder`, `QueryTracer` and `ControllerAnalyzer`. Each call re-runs `is_dir()` and, when the pattern is not a literal directory, `glob()`.

On the defaults that is two `is_dir()` calls and invisible — which is why the timings above are clean. On the layout this PR exists to serve it is not. A 90-package monolith with `packages/*/src`:

| configuration | per call | 5,000 lookups |
|---|---:|---:|
| `['app', 'src']` | 0.002 ms | 11 ms |
| `['app', 'packages/*/src']` | **0.261 ms** | **1,304 ms** |

So the change is free for applications that do not need it and costs a second-plus for the ones that do, re-globbing an unchanged directory tree thousands of times.

Memoising `resolve()` on `(root, patterns)` fixes it. I prototyped it on your branch — the glob case goes **1,304 ms → 17 ms** and all 277 tests still pass:

```php
$memoKey = $root."\0".implode("\0", $patterns);
if (isset(self::$resolved[$memoKey])) {
    return self::$resolved[$memoKey];
}
```

One caveat: a process-static memo goes stale if a directory appears during a long-lived `--watch` run. `ProjectAnalyzer::analyze()` already calls `ProjectFileIndex::clear()` for exactly that reason, so there is an established place to clear it.

## 2. An ambiguous namespaced view resolves to an arbitrary package

`firstExistingView()` prefers a root whose path carries the hint, and otherwise returns `$fallback ??= $candidate` — the first root that has a file at that relative path. Two packages with a same-named view, `packages/*/resources/views`:

```
zeta::orders.index       -> packages/zeta/resources/views/orders/index.blade.php   correct
shipping::orders.index   -> packages/alpha/resources/views/orders/index.blade.php  <- wrong package
orders.index             -> packages/alpha/resources/views/orders/index.blade.php  <- arbitrary
```

`zeta::` resolves correctly, so the hint heuristic does its job when it can. The problem is the last resort: a hint matching no root silently links to whichever root sorted first. That is a confidently wrong edge rather than a missing one, and for a graph people read to answer "what renders this?", wrong is worse than absent.

It is also a new failure mode rather than a pre-existing one — with a single views root there was never an arbitrary choice to make.

Suggestion: when a hint is present and no root matches it, return `null` if more than one root has the file. Ambiguity is exactly the case where guessing does not pay. (The unnamespaced `orders.index` line has the same shape; same fix applies, or at minimum it is worth deciding deliberately rather than by array order.)

## Smaller notes

- The `str_contains($viewRoot, '/'.$segment.'/')` segment match will accept `/billing/` for a hint of `billing-pro`. Narrow, but it is the same "close enough" step that produces the wrong edge above.
- The `IncrementalAnalyzer` containment change is a genuine fix, not just a tidy-up — that loose `str_contains($f, '/app/')` is the same bug I fixed in `ScanCommand` in #76 and again in `GraphProvenance` in #84. Good catch, and routing both through one anchored helper is the right shape.
- Base is `v2.x`, which is currently level with `main`, so nothing is stale — just be aware most recent work merged via `main`.

Finding 1 I would fix before merge; finding 2 I would at least decide deliberately. Everything else here is solid work, and the byte-identical result across three applications is the strongest evidence a change this size can carry.

webard:
Thanks - both findings were real, and I could reproduce each before changing anything.

**1. Unmemoised resolution.** Confirmed independently against a 90-package tree: `resolve()` costs 0.002 ms for `['app', 'src']` and 0.189 ms for `['app-modules/*/src']`, and it is called once per FQCN that cannot be placed. Memoised on `(root, patterns)` as you suggested - `classFilePrefixes()` goes through `resolve()`, so it is covered too. 5,000 lookups: 943 ms → 1 ms on the glob configuration, 12 ms → 0 ms on the defaults.

The memo is cleared next to `ProjectFileIndex::clear()` in `ProjectAnalyzer::analyze()`, and also at the top of `ScanCommand::collectMtimes()`. That second one goes past your suggestion deliberately: a poll is exactly when a directory may have appeared, and `collectMtimes()` runs *before* `analyze()`, so without it a new package stays invisible until some unrelated change forces a rebuild.

**2. Ambiguous view resolution.** Agreed that a confidently wrong edge is worse than a missing one. `firstExistingView()` is now `uniqueExistingView()`: it collects every root that holds the template, a hint decides only when it names exactly one of them, and otherwise a path comes back only if there is a single candidate. I applied that to the unnamespaced case as well rather than leaving it to array order - an asymmetric rule seemed harder to justify than one rule.

Your smaller note is fixed with it: hint matching is now whole path segments, with the `<vendor>-<package>` convention handled explicitly, so `billing-pro` no longer accepts `packages/billing` while `filamerce-core` still finds `app-modules/core`.

The `phpFiles()` / `contains()` realpath asymmetry now carries the note you asked for.

Nothing changes for a single views root, so the three existing `ResolveBladePath` tests pass untouched. Suite is `286 passed (901 assertions)`, PHPStan 0, Pint clean.

One question on the base. This branch needs `SourceDirectories` from #90, so I would have stacked it, but GitHub only accepts a base branch that exists in this repository - a fork branch is not offered. Leaving it on `v2.x` means the diff carries #90's commit until that merges, after which it shrinks to its own 21 files. Happy to leave it, or to retarget at `main` if you would rather these land there - you mentioned most recent work merged that way, and the two refs are currently identical.

SanderMuller (approved):
Both findings are fixed, and I re-ran the exact measurements that produced them rather than reading the diff.

## Finding 1 — glob resolution memoised

```
                          before          after
['app', 'src']            11 ms      ->     1 ms
['app', 'packages/*/src'] 1,304 ms   ->    17 ms      (5,000 lookups, 90-package tree)
```

You also wired `SourceDirectories::clear()` into both places it needs to be — `ProjectAnalyzer::analyze()` beside `ProjectFileIndex::clear()`, and `collectMtimes()` so a memo cannot survive a watch poll, which is exactly where a directory appears. `sees a directory that appeared once the memo is cleared` pins the second one.

## Finding 2 — ambiguity now returns null

Re-ran the collision probe:

```
zeta::orders.index       -> packages/zeta/...     correct, hint names one root
shipping::orders.index   -> (unresolved)          was: packages/alpha/... (wrong package)
orders.index             -> (unresolved)          was: packages/alpha/... (arbitrary)
zeta::only.here          -> packages/zeta/...     unambiguous, still resolves
only.here                -> packages/zeta/...     unambiguous, still resolves
```

The last two are the ones I was watching for — it would have been easy to fix the wrong-edge problem by giving up on multi-root resolution entirely, and you didn't. `uniqueExistingView()` reads better than what I suggested, and `viewRootCarriesHint()` closes the substring note as well: I checked `billing::` against a `packages/billing-pro/` root and it correctly declines.

## Verified again on real applications

- Full build output **byte-identical to the PR base** on all three production applications, and identical again through a symlinked project root.
- `286 passed (901 assertions)`, PHPStan 0 errors, Pint clean.
- Mutation-checked two of the new tests: remove the memo lookup and `reuses a resolved glob` dies; restore the first-match pick and `refuses to guess` dies.

On timing I can only say no regression was detected, not put a number on it — this machine is too loaded right now to resolve a small delta, and the facade phase I would normally use as a noise canary is one the bundled #90 legitimately changes. What I can say is that parse counts are identical, the memo strictly removes work, and the output is byte-identical.

## One non-blocking note

`does not accept a root whose name merely starts the hint` does not fail if `viewRootCarriesHint()` is reverted to a substring test — I tried it and the test still passes. The fixture asks whether root `alpha` answers for hint `alpha-pro`, and a substring test says no to that too. The discriminating direction is the other one: hint `alpha` against a root named `alpha-pro`, where a substring match wrongly succeeds. The implementation already gets that right; only the test does not pin it.

## Approving

Good change, and the parts I pushed back on came back better than I suggested rather than merely patched. The one note above is a test-strength nit, not a defect.

webard:
You are right, and I checked it the way you did rather than taking it on trust: mutating `viewRootCarriesHint()` to `str_contains()` leaves `does not accept a root whose name merely starts the hint` green. It pinned nothing.

The discriminating case is now in — two roots a substring test cannot tell apart, `packages/alpha` and `packages/alpha-pro`, both holding the same template, and hint `alpha`. Under whole-segment matching the hint names exactly one root and resolves it. Under a substring test it names both, stops being able to rescue the ambiguity, and the view comes back unresolved — so the test fails, which is what it is there for. Same mutation now kills the new test and leaves the old one alone, exactly as you described.

I kept the old test. It still asserts something true, just not that.

`287 passed (902 assertions)`, PHPStan 0, Pint clean.

Thanks for the review on this one — the ambiguity finding in particular. Returning a wrong edge rather than none was the kind of thing that would have sat in the graph looking authoritative for a long time before anyone questioned it.

SanderMuller (approved):
Re-verified against the new head `ae2a2b4e`. My earlier approval was recorded against `cadb5a41`, so it
described a tree that no longer exists.

The added test is the discriminating one. Reverting `viewRootCarriesHint()` to `str_contains()`:

```
✓ does not accept a root whose name merely starts the hint          still passes, as you said
⨯ matches a hint against whole path segments, not a substring       fails
```

So the property is pinned now rather than merely true. `287 passed`, and the mutation restores cleanly.

Nothing else in the branch moved. Approval stands on this head.

Separately, and to correct myself again: I have now reviewed #90 on its own, which this branch also
carries. My approval here always covered it, whatever I wrote at the time.

SanderMuller (approved):
Re-reviewed at `bfe4f26c`. My approval was recorded against `ae2a2b4e`, which two rebases have replaced, and the head moved again while I was measuring, so this describes the current one.

Most of the branch is the work I approved: the net patch is line-for-line what it was, minus #90's part, which is on `main` now. One thing is genuinely new since then, so I reviewed that rather than re-reading the rest.

## The new part: a memo on `SourceDirectories::resolve()`

A process-static cache keyed on root plus patterns. The claim it rests on is the docblock's, that it is cleared whenever the filesystem may have moved, so I ran it rather than read it:

```
first resolve        app, packages/alpha/src
new package appears  app, packages/alpha/src        <- memo still in force, as designed
after clear()        app, packages/alpha/src, packages/beta/src
```

and the case that decides whether the memo is safe, two `analyze()` calls in one process with a package created between them:

```
inside the second analyze: app, alpha/src, beta/src, delta/src, gamma/src
```

Both new packages are there, so `analyze()` really does clear it and a watch-mode rebuild cannot be served a stale directory list. `collectMtimes()` clears it too, which is the other place a poll could go stale. `ProjectFileIndex::clear()` sits on the line above in `ProjectAnalyzer`, which is the right place for both to live.

## The rest, re-run rather than re-read

- Output gate against today's `main`: byte-identical on three applications. A null arm on each came out identical too, so the corpora were stable while I measured.
- The mutation that produced my one finding still bites: replacing the segment match in `viewRootCarriesHint()` with `str_contains()` fails `it matches a hint against whole path segments, not a substring of the root`, and nothing else.
- `366 passed (1047 assertions)`, PHPStan clean, Pint clean.

## Your two red jobs are not yours

`PHP 8.2 / Laravel ^10.0` and `PHP 8.3 / Laravel ^11.0` fail on `ScopedRebuildTest > it refuses an edit that moved a call` with `ScopedRebuildNotApplicable not thrown`. Twelve jobs on the same commit pass, which is the first sign it is not the code under test.

It is a timing bug in a test helper that came in with #94:

```php
static $age = 900;
touch($path, time() - $age);
$age++;
```

Both terms grow by one per write, so a write that lands a second after the previous one repeats the previous mtime:

```
write 3: clock 1001, age 902 -> mtime 99
write 4: clock 1001, age 903 -> mtime 98
write 5: clock 1002, age 904 -> mtime 98   <- repeats
```

The two versions that test writes, `controllerCalling('run')` and `controllerCalling('log')`, are both exactly 200 bytes, and `PhpFileParser` keys its cache on `path : mtime : size`. So the second write is served the first version's AST, the moved call is invisible to the fresh build, and the check that should refuse the rebuild has nothing to refuse.

It only fires when a second elapses inside that one test, which is why it is rare and why a slower runner hits it. I could not reproduce it here in 25 runs of that file alone, 15 full-suite runs, or a replay of the seed CI failed with.

Your branch touches neither the test nor the helper. It does add tests, which shifts the timing and can make an existing flake likelier to land, and that is the whole of the connection. The fix belongs on `main`: capture the clock once, `$base ??= time()`, so the mtimes fall strictly regardless of how long the run takes. I will raise it there.

Approving the current head.

SanderMuller:
Followed this up: the two red jobs are fixed by #112, which is one line in the helper.

I proved the mechanism rather than inferring it. Putting a `sleep(1)` between the helper's writes makes the failure deterministic on `main` — the same `ScopedRebuildNotApplicable not thrown` your run reported — and with #112 in place the whole file passes under that same forced condition. Nothing in your branch is involved beyond adding tests, which shifts the timing.

Once #112 lands, a re-run of your two jobs should be green.

PR #98, opened 2026-08-31
Fix: drop subgraphs whose route no longer exists

webard:
## Drop subgraphs whose route no longer exists

A scan writes one subgraph blob per tab and never removes the ones it did not write. `subgraphIds()` reports whatever is stored, so a route that is deleted or renamed leaves its blob behind and the viewer keeps listing a tab for a surface the application no longer has.

Measured on one application: a package upgrade changed a module's route surface, and the eight URIs it dropped left **33 orphaned tab files** in `storage/app/laravel-brain/`. Three days and several scans later they were still listed — there is no path by which they ever go away, short of deleting the directory by hand.

Both drivers leaked: `FileGraphStore` never unlinked a file, `DatabaseGraphStore` never deleted a row. So pruning goes on the `GraphStore` contract, next to the `subgraphIds()` it mirrors, and both places that persist a scan call it — `ScanCommand` and `BrainController`.

### Why pruning to "what this scan wrote" is safe

This was the one thing worth checking, because getting it wrong would delete live tabs on every `--watch` poll rather than leave dead ones around.

It holds: a scoped rescan merges its partial result into the previous graph and the splitter runs on the **merged** graph, so `$result->subgraphs` is the complete tab set in both the full and the scoped mode. Both modes then persist through the same block in `ScanCommand::runScan()`. There is no path that writes a subset of tabs.

The manifest is not a tab and is excluded by `subgraphIds()` already, so it survives pruning — including `pruneSubgraphsExcept([])`, which there is a test for.

### Alternative considered

A `--prune` flag, leaving the current behaviour as the default. I did not take it: the store is a cache of the last scan, and a blob no tab can address is not data the flag would be protecting. Happy to add the flag instead if you would rather the drivers never delete without being asked.

### Tests

5 new tests against `FileGraphStore`: a subgraph the scan did not write is dropped, ones it did write are kept, the manifest is never pruned, an empty store is a no-op, and a rewritten subgraph keeps its new content.

`DatabaseGraphStore` has no test — the Unit suite runs on plain PHPUnit with no Laravel container or connection, and standing one up for this seemed a larger change than the fix. The implementation is the same query `subgraphIds()` already uses, with `delete()` and a `whereNotIn`.

Full suite: `259 passed (861 assertions)`, up from 254. `pint` clean, `phpstan` 0 errors.

### Compatibility

The contract gains a method, so a third-party `GraphStore` implementation outside this package would need to add it. Both in-tree drivers implement it, and no stored payload changes shape — the only difference is that a scan now leaves the directory (or table) holding exactly the tabs the manifest lists.

### DIFF
--- a/README.md
+++ b/README.md
@@ -392,7 +392,10 @@ With the default `file` driver, `brain:scan` writes these files to
 .graph-{tab-id}.json   — Per-route subgraph (one per route)
 ```
 
-They are regenerated on every scan and are safe to gitignore:
+Every scan rewrites the whole set and drops the subgraphs it did not write, so a tab
+disappears from the viewer when the route behind it disappears from the application.
+
+They are safe to gitignore:
 
 ```gitignore
 storage/app/laravel-brain/
--- a/src/Commands/ScanCommand.php
+++ b/src/Commands/ScanCommand.php
@@ -264,6 +264,9 @@ class ScanCommand extends Command
 
         // Both a full and a scoped rescan split the whole graph, so what was just written
         // is the complete tab set; anything else in the store outlived its route.
+        // Ordering matters: the manifest above is what the viewer indexes tabs from, and it
+        // is written unconditionally, so pruning to the same set that produced it can only
+        // remove blobs nothing can address. Moving either out of this block breaks that.
         $store->pruneSubgraphsExcept(array_map('strval', array_keys($result->subgraphs)));
 
         // The one-time support prompt flag lives on disk regardless of driver.
--- a/src/Commands/ScanCommand.php
+++ b/src/Commands/ScanCommand.php
@@ -262,6 +262,10 @@ class ScanCommand extends Command
             $store->putSubgraph((string) $tabId, $subgraph->toJson());
         }
 
+        // Both a full and a scoped rescan split the whole graph, so what was just written
+        // is the complete tab set; anything else in the store outlived its route.
+        $store->pruneSubgraphsExcept(array_map('strval', array_keys($result->subgraphs)));
+
         // The one-time support prompt flag lives on disk regardless of driver.
         $storageDir = storage_path('app/laravel-brain');
         if (! is_dir($storageDir)) {
--- a/src/Http/Controllers/BrainController.php
+++ b/src/Http/Controllers/BrainController.php
@@ -57,6 +57,8 @@ class BrainController extends Controller
             $store->putSubgraph((string) $tabId, $subgraph->toJson());
         }
 
+        $store->pruneSubgraphsExcept(array_map('strval', array_keys($result->subgraphs)));
+
         return response()->json([
             'success' => true,
             'message' => 'Project scan completed successfully.',
--- a/src/Http/Controllers/BrainController.php
+++ b/src/Http/Controllers/BrainController.php
@@ -57,6 +57,9 @@ class BrainController extends Controller
             $store->putSubgraph((string) $tabId, $subgraph->toJson());
         }
 
+        // Ordering matters: the manifest above is what the viewer indexes tabs from, and it
+        // is written unconditionally, so pruning to the same set that produced it can only
+        // remove blobs nothing can address. Moving either out of this block breaks that.
         $store->pruneSubgraphsExcept(array_map('strval', array_keys($result->subgraphs)));
 
         return response()->json([
--- a/src/Storage/DatabaseGraphStore.php
+++ b/src/Storage/DatabaseGraphStore.php
@@ -65,6 +65,18 @@ final class DatabaseGraphStore implements GraphStore
         $this->put($tabId, $json);
     }
 
+    public function pruneSubgraphsExcept(array $keep): void
+    {
+        if (! $this->ready()) {
+            return;
+        }
+
+        $this->query()
+            ->where('tab', '!=', self::MANIFEST_KEY)
+            ->whereNotIn('tab', $keep)
+            ->delete();
+    }
+
     public function subgraphIds(): array
     {
         if (! $this->ready()) {
--- a/src/Storage/FileGraphStore.php
+++ b/src/Storage/FileGraphStore.php
@@ -66,6 +66,17 @@ final class FileGraphStore implements GraphStore
         return $ids;
     }
 
+    public function pruneSubgraphsExcept(array $keep): void
+    {
+        $keep = array_flip($keep);
+
+        foreach ($this->subgraphIds() as $tabId) {
+            if (! isset($keep[$tabId])) {
+                @unlink($this->path($tabId));
+            }
+        }
+    }
+
     private function path(string $tabId): string
     {
         return $this->dir.'/.graph-'.$tabId.'.json';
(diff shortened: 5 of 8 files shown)


SanderMuller:
Reviewed and tested. This is a real leak with a correct fix, and the one thing that could have made it dangerous holds up under measurement. I have nothing I would block on.

## The safety argument holds — verified at scale

Your reasoning is that a scoped rescan splits the *merged* graph, so `$result->subgraphs` is the complete tab set and pruning to it cannot delete a live tab. That is the whole PR, so I checked it rather than read it: scoped rescan against a full scan, on three production applications, comparing tab sets.

| application | full scan | scoped rescan | tabs a prune would delete |
|---|---:|---:|---:|
| A | 367 | 367 | **0** |
| B | 423 | 423 | **0** |
| C | 537 | 537 | **0** |

Identical sets, nothing scoped-only. Same result on a synthetic three-route project. And structurally there are exactly two `putSubgraph()` callers, both writing the full set, both now pruning — no third path that writes a subset.

## The untested driver is correct — here is the test

You flagged `DatabaseGraphStore` as untested because the Unit suite has no container or connection. It does not need one: `illuminate/database`'s Capsule on `:memory:` sqlite is enough, and the whole thing is about fifteen lines. I ran your implementation against it, including the case I most expected to bite — `whereNotIn('tab', [])` compiles to `1 = 1`, which deletes *everything* it is not otherwise constrained from:

```
after writing 3 + manifest      subgraphs=["tab-a","tab-b","tab-c"]  manifest=present
prune except [tab-a, tab-c]     subgraphs=["tab-a","tab-c"]          manifest=present
prune except []                 subgraphs=[]                         manifest=present
```

Your `->where('tab', '!=', MANIFEST_KEY)` is what saves it, and the empty case matches `FileGraphStore` exactly. So the driver is right — but it is right for a reason a reader has to work out, which is the sort of thing worth pinning. Happy to hand over the harness if you want to add it.

```php
$c = new Container; Container::setInstance($c);
$capsule = new Capsule($c);
$capsule->addConnection(['driver' => 'sqlite', 'database' => ':memory:']);
$capsule->setAsGlobal(); $capsule->bootEloquent();
$c->instance('db', $capsule->getDatabaseManager());
Facade::setFacadeApplication($c);
```

## The case I expected to be a regression, and is not

A scan that finds no routes — a misconfigured `route_paths`, a wrong `--path`, a checkout mid-rebase — now empties the store. I built that and confirmed it: three stored tabs, a zero-tab scan, zero tabs afterwards.

It is not a regression, because `putManifest($result->manifestJson)` runs unconditionally a few lines earlier and already overwrote the manifest with the empty one. The viewer was already showing nothing after a bad scan; the difference is that the unreferenced blobs no longer sit on disk pretending otherwise. Worth knowing that the two are ordered that way, though — the safety of pruning depends on the manifest being written in the same block, and nothing in the code says so.

## On the `--prune` flag you offered

I would not add it, and there is precedent rather than just taste: `FileGraphStore::putManifest()` already unlinks a legacy `.graph-all.json` without being asked, with the comment "split files are canonical now". A driver deleting a blob no tab can address is established behaviour in this class. A flag would make the default the broken state.

## One small thing

`@unlink()` swallows failure, so on a directory that has become read-only the prune quietly does nothing and the leak persists with no signal. Safe direction, and I would not change the `@` — but if `subgraphIds()` still reports a tab the scan did not write on the next run, nothing will ever say why.

`259 passed`, PHPStan clean, Pint clean, all confirmed here. Good fix — the measurement on the scoped path is the part I would have wanted before merging, and it comes out clean.

webard:
Thanks — measuring the scoped path across three applications is the check I could not run myself, and it is the one the whole PR rests on. Both suggestions taken.

**The database driver is pinned now.** You were right that it needs no application: four tests on your Capsule harness, and the one that earns its place is `pruneSubgraphsExcept([])`. The comment on it says what you worked out — `whereNotIn('tab', [])` compiles to `1 = 1`, so the statement deletes every row it is not otherwise constrained away from, and `tab != __manifest__` is the only thing standing between an empty scan and losing the manifest with it. Correct before, but for a reason nothing recorded.

**The ordering is named.** Both call sites now say that the manifest is written unconditionally a few lines above and that pruning to the set which produced it can only remove blobs nothing can address — so moving either out of that block breaks the argument. That was the sharpest thing in your review: the safety property was real and entirely undocumented.

**`@unlink()` stays as it is**, and I have written down why rather than leaving it silent. A prune that fails on a read-only directory does leave the leak with no signal, but reporting it needs a logger this class does not have — that is a design change, not a note, and not one to smuggle into a bug fix.

**On `--prune`:** the `.graph-all.json` precedent settles it better than my reasoning did. I was arguing from what the store is *for*; you pointed at what this class already *does* without being asked. Dropping the offer.

`263 passed (868 assertions)`, PHPStan clean, Pint clean.

SanderMuller (approved):
Both notes are addressed, and I re-ran the probes that produced them rather than re-reading the diff.

## The database tests bite

Four tests, and I broke each thing they guard to check they fail:

| mutation | test that dies |
|---|---|
| remove `where('tab', '!=', MANIFEST_KEY)` | keeps the manifest when the keep list is empty |
| make `pruneSubgraphsExcept()` a no-op | drops the rows the scan no longer wrote |
| remove the `ready()` guard | is a no-op before the table exists |

The empty-list test is the one worth having, and the comment on it now records why the code is correct
instead of leaving the next reader to derive it.

## The central claim still holds

Re-ran the scoped-versus-full tab comparison on three production applications against this head:

| application | full scan | scoped rescan | would be deleted |
|---|---:|---:|---:|
| A | 537 tabs | 537 | 0 |
| B | 367 tabs | 367 | 0 |
| C | 423 tabs | 423 | 0 |

The database probe reproduces exactly as before: prune to two of three leaves those two, prune to an empty
list empties the subgraphs and keeps the manifest. The zero-tab scan against a populated store still ends
consistent, manifest and blobs agreeing.

## On the ordering comment

That was the sharpest thing left, and naming it is the right fix. The safety property was real and lived
nowhere — a later refactor moving `putManifest()` out of that block would have broken pruning with nothing
to catch it.

## `@unlink()`

Agreed, and the reasoning belongs in the code where you put it. Reporting the failure needs a logger this
class does not have, and adding one to a bug fix is the wrong place for that argument.

## What I did not check, and why

The full-build output gate does not apply here. This PR touches storage, the two persist call sites, the
README and tests, and nothing under `src/Analysis`, `src/Graph` or `src/Parser`, so a byte-identical result
would say nothing about it. Reporting one would be theatre.

No performance claim is made and I did not measure one. Pruning adds a directory listing per scan and
unlinks only orphans, which on a 537-tab store is not a number worth publishing.

CI has not reported on this branch at all — fork pull requests need a maintainer to approve the workflow
run. So the green below is mine, run locally against `ccd616c9`, the exact head: `263 passed (868
assertions)`, PHPStan 0 errors, Pint clean.

## Approving

Nothing blocking, and both earlier notes closed with evidence rather than assertion.

PR #99, opened 2026-08-31
Fix: breadth-first layout never terminates on a graph with a cycle

webard:
## `layoutBreadthFirst` never terminates on a cycle reachable from a root

Selecting **Breadth-first** as the layout algorithm hangs the browser tab, and with it the UI, on a large share of graphs. The graph does not have to be big: the one I first hit it on has 20 nodes and 25 edges.

The function is not breadth-first. It relaxes toward the **longest** path:

```ts
if (!level.has(v) || level.get(v)! < nextLv) {
  level.set(v, nextLv)
  q.push(v)
}
```

Every trip around a cycle finds a longer path, so the condition holds again, the node is re-queued, and the level grows without bound. There is no visited set, no cycle detection and no depth cap, so the loop cannot end — and it runs on the main thread.

### A cycle is the normal case, not an exotic one

> **Precision, from review:** it takes a cycle **reachable from a root**, not any cycle. A bare inverse pair with nothing pointing into it gives both nodes an in-degree of 1, so `roots` is empty, the queue never runs, and the old code returns immediately — with the two cards overlapping, which is #100's defect rather than this one. Everything real has a controller pointing into the models, so the two cases coincide in practice; the distinction matters only if the sentence above is read as a spec.

An inverse pair of Eloquent relations is a two-node cycle: `hasMany` on one model, the matching `belongsTo` on the other. The tab that led me here contains four of them, all ordinary:

```
AmazonCredentials --[hasMany]--> AmazonCredentialsChannel --[belongsTo]--> AmazonCredentials
AmazonCredentials --[hasMany]--> AmazonReport             --[belongsTo]--> AmazonCredentials
AmazonCredentials --[hasMany]--> AmazonNotificationSubscription --[belongsTo]--> …
AmazonNotificationDestination --[hasMany]--> AmazonNotificationSubscriptionTenant --[belongsTo]--> …
```

Running the exact loop over that tab's stored graph: **2,000,000 iterations, queue still non-empty, level counter at 285,718.**

Across one application's 623 stored tabs, **136 never terminate** — including `erd--models`, the full model ERD at 183 nodes, which is among the first things a person opens.

### The fix

Level each node once, on the first path that reaches it. That is what breadth-first means, it matches the "similar to Cytoscape breadthfirst" the docblock claims, and it terminates by construction:

```ts
let head = 0
while (head < q.length) {
  const u = q[head++]
  const lv = level.get(u)!
  for (const v of adj.get(u) ?? []) {
    if (!level.has(v)) {
      level.set(v, lv + 1)
      q.push(v)
    }
  }
}
```

All 623 tabs now complete, the worst in 130 iterations.

**This is a visible change, not only a fix.** A node whose deepest path was longer now sits at its shallowest depth instead, so some layered graphs will lay out slightly differently. That is the difference between breadth-first and longest-path layering, and the shipped name and docblock both say the former.

The `q.shift()` also became an index walk, since shifting an array is O(n) and there is no reason to pay it.

### Also in the same function

`nodes.find((n) => n.id === id)` inside the per-layer loop is O(n²) — 33k scans on the 183-node ERD. Not the hang, and I would not normally touch it in a bug fix, but it is inside the loop being repaired, so it is now a lookup map.

### Tests

There is no JS test runner in `frontend/` — no vitest, no jest — so there is nowhere to put a regression test for this, and adding a test framework is not something to smuggle into a bug fix. Happy to open that separately if you want it; a termination test for this function is exactly the kind that earns its place.

What I can offer instead is a reproduction that needs no framework. Against any stored graph JSON, the old loop does not terminate and the new one does:

```
tabs examined       623
old loop, hangs     136      (cap: 200,000 iterations)
new loop, hangs       0      (worst case: 130 iterations)
```

### Verification

- `npm run lint` clean, `tsc -b` clean.
- Bundle rebuilt: only the app chunk changed hash (`index-DSQtxLY2.js`); the css and vendor chunks hash identically to what is already committed, so the diff is one source file, one new chunk and the blade reference.
- Confirmed in the emitted bundle that the guard is now `l.has(n) || (l.set(n, t + 1), u.push(n))` and the old `l.get(n) < e` comparison is gone.


SanderMuller (approved):
Reproduced the hang, confirmed the fix, and measured the layout change you flagged. Approving.

## The termination claim, on three other applications

I ran both levelling rules over every stored tab of three production applications, building roots and
adjacency exactly as `graphLayoutD3.ts` does, with a 200,000-iteration cap:

| application | tabs | old rule hangs | new rule hangs | new worst case |
|---|---:|---:|---:|---:|
| A | 537 | 1 | 0 | 126 iterations |
| B | 367 | 1 | 0 | 204 iterations |
| C | 405 | 20 | 0 | 162 iterations |

Different magnitude from your 136 of 623, so the share clearly depends on how relation-heavy an
application is. The conclusion is the same everywhere, and on two of the three the tab that hangs is
`erd--models` — the one you singled out as among the first anyone opens.

## The visible change is real and bounded

You were right to call this a behaviour change rather than only a fix, and nobody had put a number on it.
On the tabs where the old rule terminates, so the two are comparable:

| application | tabs with any node moved | nodes moved | deepest drop |
|---|---:|---:|---:|
| A | 47 of 536 | 333 of 9,249 (3.6%) | 4 layers |
| B | 76 of 366 | 276 of 5,625 (4.9%) | 3 layers |
| C | 87 of 385 | 277 of 6,333 (4.4%) | 4 layers |

So a minority of tabs shift, and where they do it is a few percent of nodes moving up by a layer or three.
That is a redraw people may notice on a graph they know well, not a re-layout.

## What else I checked

The shipped artifact, which is what actually reaches a user: the committed chunk contains
`l.has(n)||(l.set(n,t+1),u.push(n))`, has no remaining `l.get(n)<` relax, and the blade points at it. A
source fix with a stale bundle would have looked identical in review.

Nodes that no path reaches — a component that is a pure cycle has no zero-indegree root — still fall to
layer 0 through the existing backfill above the layer map. Both rules level the same set, so that path is
unchanged.

`byId.get(id)` versus `nodes.find(...)` differs only if two nodes shared an id, and the graph dedupes on
id, so the swap is equivalent as well as faster.

The old chunk staying in the tree matches what every previous bundle bump here does. There are 240-odd
committed app chunks and none has ever been deleted, so that is a repository habit rather than anything
this PR should carry.

`composer test` on this head: 254 passed, PHPStan 0 errors, Pint clean. `tsc -b` and `eslint` both clean.

## Gates I did not apply

The full-build output gate does not apply: nothing under `src/Analysis`, `src/Graph` or `src/Parser`
changes, so a byte-identical result would say nothing. No scan benchmark either, since the cost this fixes
is browser-side and the iteration counts above measure it directly.

CI has not run on this branch at all, so the green above is mine, run locally against `7201e8e8`.

## On the missing test

Your reading is right. There is no runner in `frontend/` — the scripts are `dev`, `build`, `lint`,
`preview`, and no vitest, jest or playwright in the dependency tree — so there is nowhere to put a
termination test, and adding a framework inside a bug fix would be the wrong trade. Worth opening
separately, and this function is a good argument for it: a loop whose failure mode is an unresponsive tab
is exactly what a two-line test would have caught.

SanderMuller:

This one is harder than the others to picture from prose: the before is a browser that stops responding, so the pair to capture is the same model-heavy tab in `breadthfirst` on each side. A short recording would say more than a still, since what changed is that the page comes back at all.

webard:
Screenshots added to #100, #101 and #107. **For this one I cannot produce a useful pair, and I would rather say so than post something that looks like evidence and is not.**

The before state is a page that stops responding. A still of it is a still of the last frame that rendered — indistinguishable from a working graph, because the layout never completes and nothing new is painted. The screenshot would show the previous layout and a spinner that is not spinning. Anyone reading it would learn nothing except that I took a screenshot.

A recording has the same problem in a different shape: it would be several seconds of a motionless page. That is what the bug looks like, and it is also what a broken recording looks like.

What actually distinguishes the two states is that one terminates:

```
three nodes, root -> a -> b -> a

head   returned in 0.17ms
base   still running after 2 minutes, killed
```

and the reviewer reproduced that independently. The 2,000,000 iterations reaching level 285,718 before I stopped counting is the same fact in more detail. I do not think a picture improves on either.

If it would help a reader who has no checkout, the one thing I could capture is Chrome's own "Page unresponsive" dialog over the frozen tab. Say the word and I will, but it is closer to a photograph of a symptom than a demonstration, and the numbers above are what convinced me the diagnosis was right.

PR #100, opened 2026-08-31
Fix: layered layouts overlap because spacing ignores node size

webard:
## Before and after

Same tab, same scan, same viewport, fit to view — only the commit differs (`7201e8e` against `6f1d59f`). Driven with Playwright rather than posed by hand.

| before | after |
|---|---|
|  |  |

Cards sit on top of each other in both directions on the left: `…IsResour…`, `…entials` and `AmazonCre…` are labels a neighbouring card is covering. The rows are pitched at a constant that is narrower than the cards they carry, so the wider a card is, the further it reaches into the one beside it.

## Layered layouts space nodes by a constant pitch that ignores how big they are

> Builds on #99, which is the other defect in the same function. This branch is stacked on it so the diff stays to the spacing; merge that one first.

`layoutBreadthFirst` positions cards with `spacingX`/`spacingY = 88`, used as the distance between card **centres**:

```ts
node.x = i * spacingX - ((idsInLayer.length - 1) * spacingX) / 2
node.y = l * spacingY
```

A card is 185–270 wide (`buildLayoutNode` clamps `CARD_W_MIN`…`CARD_W_MAX`) and 90 tall (`CARD_H`). An 88px centre pitch therefore guarantees an overlap of **97–182px horizontally** and **2px vertically** — every card in a layer sitting on top of its neighbour, with the labels unreadable.

### It is not a consequence of #99

Worth stating plainly, because the two defects live in the same function. The positioning block is untouched by the levelling fix — `git diff` over that PR shows no `spacingX`, `spacingY`, `node.x` or `node.y` line.

And it is measurable on graphs the old code could always lay out. Of one application's 623 stored tabs, 444 are acyclic, so the old levelling loop terminated on them and they rendered: **all 444 came out with overlapping cards.** The overlap was invisible on the cyclic ones only because those never finished laying out at all.

### The fix

Measure the gaps against what they have to clear. Within a layer each card contributes its own width (or its own height, when the layout runs left-to-right), and layers are separated by the tallest card in the one before:

```ts
const rowWidth = layerNodes.reduce((sum, n) => sum + n.width, 0) + gapWithinLayer * (layerNodes.length - 1)
const rowHeight = Math.max(...layerNodes.map((n) => n.height))
let x = -rowWidth / 2
for (const node of layerNodes) {
  node.x = x + node.width / 2
  node.y = cursor + rowHeight / 2
  x += node.width + gapWithinLayer
}
cursor += rowHeight + gapBetweenLayers
```

The parameters change meaning — from centre pitch to the gap between card edges — so they are renamed to say so, and default to 60/110 to match the `nodesep`/`ranksep` already handed to dagre. The two layouts read alike now instead of one of them being unusable.

Re-measured over the same tabs after the change: **0 with overlapping cards.**

### `layoutGrid` has the same defect, milder

A fixed `cellW = 200` against cards up to 270 wide overlaps by 70px. Same root cause — a constant chosen without reference to the thing it has to fit — so the cell is derived from the widest card in the graph. Its call site passed `200, 130` as cell sizes; it passes nothing now and takes the gap defaults.

I would normally leave a second function out of a bug fix, but this is the identical mistake ten lines away, and fixing one while leaving the other reads as an oversight rather than a decision.

`layoutDagre` was already correct: it hands real `width`/`height` to dagre and lets it separate them. `layoutCircle` and `layoutForce` position by radius and by simulated collision, and neither uses a fixed pitch.

### Verification

- Measured geometrically, using the same `buildLayoutNode` width formula the renderer uses: 444/444 tabs overlapping before, 0/580 after.
- Confirmed in a browser against a real application: cards separated, edge labels (`hasMany`, `belongsTo`) legible between them where they were previously covered.

No test — the same reason as #99: `frontend/` has no JS test runner, and adding one belongs in its own change rather than in a bug fix. The geometric check above is reproducible against any stored graph JSON.


SanderMuller (approved):
Measured the overlap before and after, across three production applications and all three layouts you
touched. Approving.

## Cards did overlap, on nearly every tab

Laying every stored tab out both ways, using the renderer's own width formula
(`clamp(185, 270, longest * 7.6 + 44)`, height 90):

| application | tabs | overlapping before | after |
|---|---:|---:|---:|
| A | 537 | 536 | 0 |
| B | 367 | 362 | 0 |
| C | 405 | 400 | 0 |

The worst single tab had 1,345 overlapping card pairs. The few that came out clean are the ones with a
single node per layer, where an 88px pitch has nothing to collide with.

That is a stronger result than your 444 of 623, because it does not depend on the tab being acyclic: I
levelled with the fixed rule this branch already carries, so every tab lays out and every tab can be
measured.

## The two branches and the grid, separately

You reported the aggregate. Splitting it, since `LR` is a separate code path and `layoutGrid` is a second
function:

| application | TB before → after | LR before → after | grid before → after |
|---|---|---|---|
| A | 536 → 0 | 536 → 0 | 536 → 0 |
| B | 362 → 0 | 362 → 0 | 358 → 0 |
| C | 400 → 0 | 400 → 0 | 400 → 0 |

The grid was overlapping about as widely as the layered layouts, so pulling it into this change was the
right call rather than scope creep.

## Your independence claim holds

`git show` over #99 touches no `spacingX`, `spacingY`, `node.x` or `node.y` line; #100 touches fourteen.
The two defects share a function and nothing else.

## Also checked

**Both changed signatures have exactly one caller.** `layoutBreadthFirst` is called with three arguments,
so the renamed defaults apply cleanly, and `layoutGrid`'s `200, 130` is correctly dropped at its only call
site. Nothing is left passing the old numbers into parameters that now mean something else, which was the
risk worth checking in a rename like this.

**`Math.max(...layerNodes.map(...))` is safe here.** The widest layer across all three applications is 86
nodes, far from anything a spread has trouble with.

**The bundle is what the source builds.** `npm run build` reproduces the committed `index-odlQ6dFl.js` byte
for byte, with no other asset moving.

`composer test` on this head: 254 passed, PHPStan 0, Pint clean. `tsc -b` and `eslint` clean.

## Gates I did not apply

No output gate: nothing under `src/Analysis`, `src/Graph` or `src/Parser` changes. No scan benchmark, since
the cost is browser-side. CI has not run on this branch, so the green above is mine, local against
`6f1d59f6`.

No test, for the reason you and #99 both give.

## One correction of my own

I approved #101, which sits on top of this branch, and wrote that my approval covered only its own commit
and that this one should land on its own review. That was wrong about how GitHub works: an approval applies
to the whole diff, so I had already approved this commit before reading it. Reviewing it now is the right
way round, and it holds up — but the sequencing was mine to get right and I did not.

SanderMuller:

The same tab in `dagre`, `breadthfirst` and `grid` on both sides would carry it. Overlapping cards are obvious in a picture and have to be taken on trust from a table of pitch-versus-width numbers.

webard:
Added to the description — a before/after pair at the top.

Same tab, same scan data, same viewport, fit to view, driven with Playwright so the two frames come out of one run and differ only by the commit (`7201e8e` against `6f1d59f`).

The before is more legible than the numbers were: `…IsResour…`, `…entials` and `AmazonCre…` are labels with another card sitting on top of them. A constant row pitch narrower than the cards it carries means the wider a card is, the further it reaches into its neighbour, which is hard to state and obvious to look at.

PR #101, opened 2026-08-31
Fix: a component with no entry point is flattened onto level 0

webard:
## Before

Four Apilo models, all reached only by each other: `ApiloCredentials ⇄ ApiloMigrationRun ⇄ ApiloMigrationStep`, `ApiloMigrationRun ⇄ ApiloRecord`. Every one has an in-degree of at least one, so the component has no root, the queue never reaches it, and the four come out in a flat row with the hub wedged between its own children.

Their positions, read off the laid-out nodes on each side:

```
before (6f1d59f)   (-6518, -139)  (-6273, -139)  (-6028, -139)  (-5783, -139)
after  (b7148f7)   (-5779, -195)  (-2754,    5)  (-2920,  205)  (-2675,  205)
```

One y before, three after — that is the whole fix, and it is more legible as four numbers than as a picture.

**There is no matching "after" screenshot, deliberately.** Once the component is levelled it is placed within the graph's own layers rather than in a row of its own, and those four nodes end up 3,104px apart across a 183-node ERD. Framing them makes a 6% zoom where nothing is readable, so a paired image would show less than the coordinates above. Whether that placement is the right one is a fair question, but it is a layout question rather than evidence this one is unfixed.

This is also the only tab of 621 in the application I measured that has a rootless component at all, which is why the defect survived: it is invisible unless a model cluster happens to have no controller pointing into it.

## A component with no entry point loses its structure

> Builds on #100, which is in the same function. Stacked so the diff stays to this change; merge that one first.

`layoutBreadthFirst` levels outward from nodes with an in-degree of 0, then sweeps up whatever the walk missed:

```ts
for (const n of nodes) if (!level.has(n.id)) level.set(n.id, 0)
```

Not every component has such an entry. Two models joined only by an inverse relation — `hasMany` one way, `belongsTo` back — form a component with nothing pointing in from outside it, so there is no node to start walking from. Every member lands on level 0 no matter what it connects to, in one undifferentiated row next to the genuine roots.

**Measured on a real graph, not a constructed one.** In one application's model ERD, **53 of 183 nodes** are never reached and get flattened. Across the same application's 580 stored graphs it is the only one affected, which is what you would expect: a route graph always has the route itself as an entry, and only the ERD is pure model-to-model.

### The fix

After the walk from the natural roots, keep seeding from the least depended-on node that still has no level, until every node has one:

```ts
for (;;) {
  let seed: LayoutNode | null = null
  for (const n of nodes) {
    if (level.has(n.id)) continue
    if (seed === null || indeg.get(n.id)! < indeg.get(seed.id)!) seed = n
  }
  if (seed === null) break
  level.set(seed.id, 0)
  q.push(seed.id)
  drain()
}
```

In-degree is a heuristic for "most root-like", which is all it has to be — inside a cycle there is no correct answer, only a more useful starting point than "all of them at once".

On that ERD the widest layer goes from **110 nodes to 75**: the 53 stop being a flat row and take their component's own depth.

That figure is application-specific and should not be read as typical. It scales with how many entry-less components a model graph happens to have, and that varies a lot — measured on three other applications the counts were 7 and 4 flattened nodes, with the widest layer going 86 → 82 on one and not moving at all (67 → 67) on another. The direction is consistent; the size of the win is not.

It also covers the degenerate whole-graph case, where no node has an in-degree of 0 at all. The queue then started empty, the loop ran zero times and the sweep put every node on level 0. No stored graph I have is in that shape, but it reproduces on eight models wired in a ring of inverse relations, where the layering goes from one flat row of eight to the eight-deep chain it actually is.

### What this does not do

It does not make a 183-node ERD readable. Seventy-five cards on one level is a very wide row whatever the spacing, and the layout that would handle that graph better is dagre — which is exactly what `pickLayoutKind` substitutes away above 80 nodes. This change stops the layout discarding structure it already has; it does not add structure that is not there.

### Verification

- Layer assignment measured against the stored graphs, using the same levelling the renderer runs: 53 flattened nodes before, 0 after; widest layer 110 → 75.
- Rendered in a browser against the real ERD: no console errors, graph draws.

No test, for the reason given in #99 and #100: `frontend/` has no JS test runner, and adding one belongs in its own change.


SanderMuller (approved):
Reproduced the flattening, measured the fix, and tested the half you said was reasoned rather than
measured. Approving, with one number worth setting expectations on.

## The flattening is real, and only the ERD

Reproducing the levelling over every stored tab of three production applications, before and after:

| application | tabs | tabs with flattened nodes | worst tab |
|---|---:|---:|---|
| A | 537 | 1 | `erd--models`, 7 of 133 nodes |
| B | 367 | 0 | — |
| C | 405 | 1 | `erd--models`, 4 of 164 nodes |

Your reasoning about why holds exactly: a route graph always has the route as an entry, so only the pure
model-to-model graph is affected, and it is the ERD every time.

## The size of the win is application-specific

Yours goes 53 of 183 flattened and a widest layer of 110 down to 75. Mine are much smaller: 7 and 4 nodes,
and the widest layer goes 86 to 82 on one and does not move at all on the other, 67 to 67. The four nodes
take their own depth but they were not what made that row wide.

So the fix is right and the direction is consistent, but a maintainer reading 110 to 75 should not expect
that everywhere. It scales with how many entry-less components a model graph has, and that varies a lot.

## The degenerate case, now measured

You flagged the whole-graph case as reasoned. It reproduces: eight models in a ring of inverse relations,
so no node has an in-degree of 0.

```
old:  L0:8                                        one flat row
new:  L0:1 L1:1 L2:1 L3:1 L4:1 L5:1 L6:1 L7:1     the chain it actually is
```

## Also checked

**The outer loop terminates.** Each pass either levels at least one node or breaks, so it runs at most once
per node.

**The seed scan is quadratic, and it does not matter.** Rescanning every node per component is O(n²) in the
number of entry-less components. The real worst case is many small ones — 400 disjoint two-cycles, 800
nodes, 400 outer passes — and that lands at 5.3 ms. My first attempt at that probe used isolated nodes,
which was wrong: they have an in-degree of 0, so they are all seeded in the first pass and the outer loop
never runs at all.

**The bundle is what the source builds.** Running `npm run build` in the branch reproduces the committed
`index-pWcB9eEi.js` byte for byte, same name and same content, with no other asset changing. That is worth
more than reading the minified output, and I would suggest it as the check for any frontend change here.

`composer test` on this head: 254 passed, PHPStan 0, Pint clean. `tsc -b` and `eslint` clean after
`npm ci`.

## Worth stating, since it raises the stakes

The default layout is `dagre`, `LARGE_GRAPH_THRESHOLD` is 80, and `pickLayoutKind` substitutes
breadth-first in above it. So every graph over 80 nodes renders with this algorithm on the default setting,
for someone who never chose it — 3, 5 and 20 tabs across these three applications. That applies to #99 as
much as to this one.

## Scope of this approval

I reviewed `b7148f77`. The branch also carries `7201e8e8` from #99, which I approved separately, and
`6f1d59f6` from #100, which I have not reviewed — that one should land on its own review rather than ride
in on this approval.

## Gates I did not apply

No output gate: nothing under `src/Analysis`, `src/Graph` or `src/Parser` changes, so a byte-identical
result would say nothing. No scan benchmark, since the cost is browser-side. CI has not run on this branch,
so the green above is mine, local against `b7148f77`.

No test, for the reason you and #99 both give: `frontend/` has no runner, and adding one belongs in its own
change.

SanderMuller:
Correcting the scope paragraph in my approval above.

I wrote that it covered `b7148f77` only and that #100 should land on its own review rather than ride in on it. That is not how a GitHub approval works — it applies to the whole diff of the PR, so approving this one already approved #100's commit, whatever the sentence said. Disclaiming a commit inside a branch I approved does not remove it from what I approved.

I have now reviewed #100 properly and approved it on its own merits: card overlap measured on three production applications, 536/362/400 tabs overlapping before and 0 after, across the TB path, the LR path and `layoutGrid` separately. So the approval here stands, and the commit I disclaimed holds up. The order was wrong, not the outcome.

webard:
Thanks — three things taken.

**The 110 → 75 framing was too strong, and the body now says so.** Your 7 and 4, with one widest layer not moving at all, is the useful counterweight: the direction is consistent, the size of the win is not, and it scales with how many entry-less components a model graph happens to have. A maintainer reading only my number would have expected the wrong thing.

**The degenerate case is measured now, by you rather than by me** — eight models in a ring, one flat row of eight becoming the eight-deep chain. The body no longer calls that half reasoned.

**Reproducing the bundle by rebuilding is a better check than the one I used**, and I am adopting it. I had grepped the minified output for the shape I expected to find, which confirms the change is present but not that the committed chunk is what this source produces. Byte-for-byte from a clean `npm ci && npm run build` answers the question I was actually asking.

On the threshold: agreed that it raises the stakes rather than being a separate defect. I had gone looking for it as one, and stopped when I found `Toolbar.tsx` already discloses the substitution — `⚠ large` carries the tooltip "Large graph: dagre auto-switched to breadthfirst". So it is documented, just not at the select where the label still reads Hierarchical. Not something I would open a PR about on its own; happy to be told otherwise.

And you are right about the scope. `6f1d59f6` is #100 and rides along only because these are stacked on the same function — it should land on its own review, not on this approval.

SanderMuller:

Pick a tab with a component that has no entry point, in `breadthfirst`, and capture it both ways. The before is the one worth showing: everything unreachable collapsed into one row beside the real roots, which is hard to describe and obvious to look at.

webard:
Added — with a caveat I would rather state than paper over.

The **before** is in the description and it is the clear half: the four Apilo models come out in a flat row, `ApiloMigrationRun` wedged between the two children that point at it. They are reached only by each other, so the component has no root, the queue never gets there, and nothing levels them.

**There is no matching after image, deliberately.** Once levelled, those four are placed within the graph's own layers rather than in a row of their own, and they end up 3,104px apart across a 183-node ERD — framing them gives a 6% zoom where nothing is readable. The coordinates say it better:

```
before   (-6518, -139)  (-6273, -139)  (-6028, -139)  (-5783, -139)
after    (-5779, -195)  (-2754,    5)  (-2920,  205)  (-2675,  205)
```

One y before, three after. A picture of that would be worse than the four numbers.

Worth knowing for anyone reproducing it: this is the **only tab of 621** in the application I measured that has a rootless component at all. That is why it survived — it is invisible unless a model cluster happens to have no controller pointing into it.

PR #102, opened 2026-08-31
Fix: report memory exhaustion instead of exiting silently, and allow a configured limit

webard:
## A scan that runs out of memory says nothing at all

On a large application `brain:scan` ends like this:

```
 ○ Scanning routes..    ✓ routes.. 380 routes (0.01s)
 ○ Scanning middleware..    ✓ middleware.. (0.00s)
 ○ Analyzing controllers..    ✓ controllers.. 223 controllers (0.21s)
 ○ Tracing full lifecycle..
```

Nine lines, stopping mid-step, exit code 255. No mention of memory, no PHP fatal, nothing in the Laravel log, empty stderr. It reads like the scanner gave up for no reason.

It ran out of memory. Measured on the same project: `--memory-limit=1024M` exits 255, `1536M` and `2048M` complete. And the reason nothing is printed is the same reason it died — with the heap full there is nothing left to allocate a report with, so PHP's own "Allowed memory size … exhausted" never renders either.

The default is `1024M`, and `normalizeMemoryLimit()` enforces `1024M` as the **minimum** the command will accept. So the value that fails is the lowest one available, the user cannot go lower, and nothing in the output suggests going higher.

### Saying what happened

The fatal is still readable at shutdown through `error_get_last()`. What is missing at that point is memory to act on it, so the command holds 256KB back and releases it first:

```php
$reserve = str_repeat(' ', MemoryExhaustionNotice::RESERVE_BYTES);

register_shutdown_function(function () use (&$reserve, $limitInEffect): void {
    $reserve = null;
    $notice = MemoryExhaustionNotice::forLastError(error_get_last(), $limitInEffect);
    if ($notice !== null) {
        $this->newLine();
        $this->error($notice);
    }
});
```

Which turns the silence into:

```
The scan ran out of memory at --memory-limit=1024M. Raise it (--memory-limit=2048M),
lift it entirely (--memory-limit=-1), or set `memory_limit` in config/laravel-brain.php
so every scan of this project gets the larger value.
```

The suggested value is double what did not fit, keeping the unit it was given, so the advice is a command to run rather than an invitation to guess.

A side effect worth knowing: with the reserve released, **PHP's own fatal renders too**. Before the change that run printed nothing; after it, the same run prints `Allowed memory size of 1073741824 bytes exhausted` as well as the line above. The reserve is not only for our message — it is what lets any error reporting work at that moment.

**And it depends on how the heap fills, which is why this looks intermittent.** Measured at a 32M limit, varying the size of the allocation that finally fails:

| failing allocation | without the reserve | with the reserve |
|---|---|---|
| 1 MB | printed | printed |
| 100 KB | printed | printed |
| 1 KB | **silent** | printed |
| 64 B | **silent** | printed |

A run that dies asking for one large block leaves that much slack behind and would have reported itself anyway. A run that dies in small increments has nothing left — and small increments is what parsing does, which is why the scan is the case that goes quiet. (Boundary measured by @SanderMuller in review; I had claimed the silence unconditionally.)

### Setting it once per project

The limit now comes from config, so an application that needs 2G stops relying on everyone remembering the flag:

```php
// config/laravel-brain.php
'memory_limit' => env('LARAVEL_BRAIN_MEMORY_LIMIT', '1024M'),
```

`--memory-limit` still wins for a single run. Omitted, the option falls through to the config, which defaults to the previous `1024M` — so nothing changes for anyone who does not set it. The signature default moved from `1024M` to empty for that reason; that is the only way to tell "not passed" from "passed the same value".

### Why the message is its own class

`MemoryExhaustionNotice` is separated from the command so the decision can be tested without exhausting memory inside a test. 7 tests: the message names the limit and the config key, the suggestion doubles, the unit is preserved, an unparseable limit still yields something concrete, and it stays silent for a normal exit, for an unrelated fatal, and for a warning that merely mentions memory. That last one matters — claiming a memory problem for an unrelated crash sends the reader the wrong way.

### Verification

- Reproduced against a real application, before and after, at the default limit: 9 lines and silence before; the same run with the notice and PHP's own fatal after. Exit code stays 255 either way, which is correct — this reports the failure, it does not swallow it.
- `254 → 261 passed`, PHPStan 0 errors, Pint clean.

PHPStan caught the first shape of this, where the reserve was a property: `never read, only written`. It was right — the reserve's lifetime belongs to the shutdown handler, not to the command object, so it is a local captured by reference now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -331,6 +331,27 @@ Stress testing uses [`laramint/laravel-stress`](https://github.com/LaraMint/lara
 
 Single-label hostnames (Docker service names with no dots) and private IPv4 ranges are automatically allowed by the host validator, so pointing the Base URL at your internal network address will work without any extra config.
 
+## Memory
+
+A scan holds the whole graph and a parsed-AST cache at once, so a large application needs more than the 1024M default. When it does not fit, PHP kills the process — and with nothing to allocate, neither PHP's own error report nor anything else can render, so the run ends mid-step at exit 255 with no output at all.
+
+The scan holds a small reserve back for exactly that moment: it releases it at shutdown and says what happened and what to change.
+
+```
+The scan ran out of memory at --memory-limit=1024M. Raise it (--memory-limit=2048M),
+lift it entirely (--memory-limit=-1), or set `memory_limit` in config/laravel-brain.php
+so every scan of this project gets the larger value.
+```
+
+Set it once per project rather than remembering the flag:
+
+```php
+// config/laravel-brain.php
+'memory_limit' => env('LARAVEL_BRAIN_MEMORY_LIMIT', '2048M'),
+```
+
+`--memory-limit` still overrides it for a single run.
+
 ## Storage Driver
 
 Scan output (the graph) can be persisted in one of two ways, selected with the
--- a/config/laravel-brain.php
+++ b/config/laravel-brain.php
@@ -63,6 +63,22 @@ return [
     ],
 
     // -------------------------------------------------------------------------
+    // Memory Limit
+    // -------------------------------------------------------------------------
+    // Memory the scan runs with, applied before analysis starts. Accepts the PHP
+    // notation (1024M, 2G) or -1 for no limit; the minimum accepted is 1024M.
+    //
+    // A scan holds the whole graph plus a parsed AST cache, so a large application
+    // needs more than the default. When it does not fit, PHP kills the process — the
+    // scan reports that and names this setting rather than exiting silently.
+    //
+    // `--memory-limit` on the command overrides this for a single run.
+    //
+    // Override via the LARAVEL_BRAIN_MEMORY_LIMIT env variable.
+    //
+    'memory_limit' => env('LARAVEL_BRAIN_MEMORY_LIMIT', '1024M'),
+
+    // -------------------------------------------------------------------------
     // Auto-Discover Routes
     // -------------------------------------------------------------------------
     // When true, RouteAnalyzer skips AST parsing of route_paths and instead
--- a/src/Commands/MemoryExhaustionNotice.php
+++ b/src/Commands/MemoryExhaustionNotice.php
@@ -1,0 +1,65 @@
+<?php
+
+declare(strict_types=1);
+
+namespace LaraMint\LaravelBrain\Commands;
+
+/**
+ * Turns PHP's memory-exhaustion fatal into something a person can act on.
+ *
+ * When the scan runs past `memory_limit`, PHP kills the process with exit code 255. On a
+ * CLI with `display_errors` off nothing reaches the terminal at all: the progress line
+ * stops mid-step and there is no message, no exit reason, and nothing in the log. The
+ * scan looks like it stopped for no reason, and the one thing that would fix it — a
+ * larger limit — is the one thing the output does not mention.
+ *
+ * The fatal is still recoverable at shutdown: {@see error_get_last()} holds it, and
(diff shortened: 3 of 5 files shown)


SanderMuller (approved):
Tested the reserve against real exhaustion rather than taking it on trust, and it earns its place. Approving.

## The reserve is load-bearing, and here is where

Your claim is that with the heap full there is nothing left to build a message with, so the reserve is what
lets anything render. I ran the pattern both ways at a 32M limit, filling the heap with different
allocation sizes:

| failing allocation | notice without the reserve | with the reserve |
|---|---|---|
| 1 MB | printed | printed |
| 100 KB | printed | printed |
| 1 KB | **silent** | printed |
| 64 B | **silent** | printed |

So the reserve matters exactly when the heap fills in small increments, which is what parsing does. A run
that dies on one large allocation leaves that much slack behind and would have printed anyway — which is
probably why this looks unnecessary until you hit it. Worth knowing the boundary, because it explains why
the bug is intermittent-looking rather than absolute.

Exit code stays 255 in both arms, as you said.

## The guards hold

Both refusals are the interesting part of the class, since claiming a memory problem for an unrelated crash
sends the reader the wrong way. I broke each:

| mutation | test that dies |
|---|---|
| drop the `E_ERROR` type check | says nothing for a warning that merely mentions memory |
| drop the `Allowed memory size of` check | says nothing for a fatal it cannot explain |

## The option fallback does what the compatibility note claims

Moving the signature default from `1024M` to empty is the part that could quietly change behaviour, so I
checked it against the input definition rather than reasoning about it:

```
flag omitted           option() = NULL      falls through to config
--memory-limit=2048M   option() = '2048M'   wins
--memory-limit=        option() = ''        does not fall through
```

The first is the one that matters and it holds. The third goes to `normalizeMemoryLimit('')` and errors on
the format, which is what it did before this change too, so it is not a regression.

## Checked and fine

I went looking for a gap between the config key and the existing floor, since the README sells it as "set
it once per project" and `normalizeMemoryLimit()` rejects anything under 1024M outright rather than
clamping. The config file comment already says `the minimum accepted is 1024M`, which is where someone
setting it would read it.

## Gates

No output gate: this touches the command, the config and a new class, and nothing under `src/Analysis`,
`src/Graph` or `src/Parser`, so a byte-identical result would say nothing about it. No benchmark either,
since holding 256KB is not a cost worth measuring against a scan that allocates hundreds of megabytes.

`composer test`: 261 passed, PHPStan 0, Pint clean. CI has not run on this branch, so that green is mine,
local against `96d9590c`.

One aside: your note that PHPStan caught the property version as "never read, only written" is a good
argument for the local-captured-by-reference shape, and it is the kind of thing worth leaving in the commit
message where you put it.

webard:
The allocation-size boundary is the part I did not have, and it is better than what I wrote.

I claimed the silence unconditionally. Your table says it depends on the size of the block that finally fails: a run dying on 1 MB leaves that much slack and reports itself anyway, a run dying on 1 KB or 64 B has nothing left. Small increments is what parsing does, which is why the scan is the case that goes quiet — and why this looks intermittent rather than absolute to anyone who meets it twice. The description carries the table now, credited.

Both refusals mutation-tested is the check I would have wanted, and the option-fallback table too — `--memory-limit=` giving `''` and erroring on format is the case I did not think to try. Agreed it is not a regression, since it did that before as well.

On the floor: you are right that the config comment is where someone setting the key would read it, and it says the minimum. If `normalizeMemoryLimit()` clamped instead of erroring I would argue for saying so louder, but rejecting outright is at least loud by construction.

PR #103, opened 2026-08-31
Fix: AI context export contains no source, and prints provenance as PHP

webard:
## The AI context export contains no source, and prints provenance inside a ```php fence

`ContextExporter` builds its budget-gated snippets from `data['source']`:

```php
$source = (string) ($node['data']['source'] ?? '');
if ($source === '') {
    continue;
}
```

No analyzer ever puts code there. The only writer of that key is `GraphBuilder`'s command node, and what it stores is `ConsoleCommandDefinition::$source` — *where the definition was found*, one of `'class'`, `'route'` or `'kernel'`.

On one application's 10,142 graph nodes the effect splits two ways. For 10,029 the key is absent, so the loop `continue`s and the export has no code in it at all. For the remaining 113 — the commands — it produces this:

````
## Source: pim:prune-orphaned-product-listings {--tenant=*} (focal)
```php
class
```
````

And because no snippet is ever emitted, `--budget` governs nothing. The same route exported at `--budget=300` and at `--budget=50000` came out byte-identical apart from the header line, and **87% of it was the two dependency tables**.

### Reading the file the node already names

Nodes carry `data['file']` — 8,454 of those 10,142 — which is the same path the viewer's Source tab loads through `BrainController::source()`. The snippets read that instead.

Measured on the same route export:

| | before | after |
|---|---|---|
| total | 5,058 chars | 21,764 chars |
| source | none | 77% |
| dependency tables | 87% | 20% |

And the budget bounds it monotonically for the first time:

```
budget    500 ->  1,264 tokens,  0 source sections
budget  2,000 ->  1,935 tokens,  1
budget  6,000 ->  5,441 tokens,  4
budget 20,000 -> 14,174 tokens, 19
```

Reading is confined to the project being analysed. Node paths come from the scan and should already be inside it, but an export is something a person hands to a model, and reading outside the tree is not a mistake worth leaving reachable.

### Renaming the key

The command node's `source` becomes `definedIn`, so nothing is called `source` that is not source. Nothing consumed it: the viewer distinguishes an edge from a node by testing `source` **and** `target` together, which a command node never satisfied — close, but it never fired.

### Tests

4 tests covering the decision: a node whose file exists yields its code under a `## Source:` heading, a missing file yields nothing, a file outside the project is refused, and a node naming no file yields nothing.

### Verification

- Reproduced against a real application before and after, including the `class`-in-a-php-fence output, which is what sent me looking.
- `254 → 258 passed`, PHPStan 0, Pint clean.

### One thing this does not fix

The two dependency tables are still emitted unconditionally, ahead of any snippet, and subtracted from the budget rather than fitted into it. At `--budget=500` the export is 1,264 tokens — 2.5× the budget — of which 87% is those tables and none is code. That is a separate defect and I will raise it on its own.

### DIFF
--- a/src/Ai/ContextExporter.php
+++ b/src/Ai/ContextExporter.php
@@ -312,11 +312,24 @@ class ContextExporter
         $sourceParts = [];
         $nodesForSource = $ctx->nodes;
 
+        // Nodes name files far more often than files exist: a class and its methods, a resource
+        // and its pages, all carry the same path. Emitting per node would put one file's whole
+        // contents in the export once for every node that names it — and every copy is spent
+        // out of the budget the rest of this method exists to respect.
+        $emitted = [];
+
         foreach ($nodesForSource as $i => $node) {
-            $source = $this->readSource((string) ($node['data']['file'] ?? ''));
+            $file = (string) ($node['data']['file'] ?? '');
+            if ($file !== '' && isset($emitted[$file])) {
+                continue;
+            }
+
+            $source = $this->readSource($file);
             if ($source === '') {
                 continue;
             }
+
+            $emitted[$file] = true;
 
             $isFocal = $i === 0;
             $label = (string) $node['label'];
--- a/src/Ai/ContextExporter.php
+++ b/src/Ai/ContextExporter.php
@@ -303,12 +303,17 @@ class ContextExporter
         $structural = implode("\n", $parts);
         $charBudget -= strlen($structural);
 
-        // Source snippets (budget-gated, focal node first)
+        // Source snippets (budget-gated, focal node first).
+        //
+        // Read from the file each node names. This used to read `data['source']`, which no
+        // analyzer ever fills with code — the only writer put a command's provenance there —
+        // so the whole budget-gated path emitted nothing on 8,454 of one project's 10,142
+        // nodes and a ```php block containing the word "class" on the remaining 113.
         $sourceParts = [];
         $nodesForSource = $ctx->nodes;
 
         foreach ($nodesForSource as $i => $node) {
-            $source = (string) ($node['data']['source'] ?? '');
+            $source = $this->readSource((string) ($node['data']['file'] ?? ''));
             if ($source === '') {
                 continue;
             }
@@ -343,6 +348,30 @@ class ContextExporter
         }
 
         return $structural.implode("\n", $sourceParts);
+    }
+
+    /**
+     * The contents of a file a node names, or '' when there is nothing safe to read.
+     *
+     * Confined to the project being analysed: node paths come from the scan, but an export is
+     * something a person hands to a model, and reading outside the tree is not a mistake worth
+     * being able to make.
+     */
+    private function readSource(string $file): string
+    {
+        if ($file === '' || ! is_file($file) || ! is_readable($file)) {
+            return '';
+        }
+
+        if ($this->projectPath !== '') {
+            $root = realpath(rtrim($this->projectPath, '/'));
+            $real = realpath($file);
+            if ($root === false || $real === false || ! str_starts_with($real, $root.DIRECTORY_SEPARATOR)) {
+                return '';
+            }
+        }
+
+        return (string) file_get_contents($file);
     }
 
     // ── JSON serializer ───────────────────────────────────────────────────────
--- a/src/Graph/GraphBuilder.php
+++ b/src/Graph/GraphBuilder.php
@@ -2025,7 +2025,10 @@ class GraphBuilder
                     'description' => $cmd->description,
                     'class' => $cmd->class,
                     'file' => $cmd->file,
-                    'source' => $cmd->source,
+                    // Where the definition was found ('class' | 'route' | 'kernel'). Not the
+                    // command's code: `source` read as source is what put the literal string
+                    // "class" into AI context exports inside a ```php fence.
+                    'definedIn' => $cmd->source,
                     'flowSteps' => $flowSteps,
                     'metrics' => $metrics ?: null,
                     'hasN1' => $hasN1,
--- a/tests/Unit/ContextExporterSourceTest.php
+++ b/tests/Unit/ContextExporterSourceTest.php
@@ -1,0 +1,79 @@
+<?php
+
+use LaraMint\LaravelBrain\Ai\ContextExporter;
+use LaraMint\LaravelBrain\Storage\FileGraphStore;
+
+/** A project with one node whose `file` points wherever the caller says. */
+function exporterFixture(string $nodeFile): array
+{
+    $project = sys_get_temp_dir().'/lb_ctx_'.bin2hex(random_bytes(6));
+    mkdir($project.'/storage', 0o777, true);
+
+    $store = new FileGraphStore($project.'/storage');
+    $store->ensureSchema();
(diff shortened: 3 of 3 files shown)

### CODE REVIEW

**Review by SanderMuller — APPROVED** [review]
Reviewed at `08496eb2`. I rebuilt the branch, ran the output gate on three applications, reproduced the before/after on two of them independently, probed the containment check against the shapes that usually defeat one, and mutation-checked all four new tests. Everything load-bearing here holds. Two non-blocking notes at the end, neither of which should hold the PR.

## The claim everything rests on

That `data['source']` never holds code and `data['file']` does. Rather than read for it, I ran the exporter on both arms against two applications.

Full-project export, budget 100,000:

| | app A before | app A after | app B before | app B after |
|---|---:|---:|---:|---:|
| export | 112,263 chars | 400,038 | 89,271 chars | 400,060 |
| in ```php fences | 538 chars | 297,300 | 168 chars | 311,158 |
| code share | 0% | 74% | 0% | 78% |

538 characters spread over 88 fences is the `class`-in-a-fence output you describe, and it is the entire code content of a whole-project export on that application. Your reading of it was right.

The monotonicity claim reproduces too. One route on app B, both arms, same store:

```
budget     before                      after
   500     2,401 tokens, 0 sections    2,401 tokens, 0 sections
 2,000     2,401 tokens, 0 sections    2,401 tokens, 0 sections
 6,000     2,401 tokens, 0 sections    6,008 tokens, 1 section
20,000     2,401 tokens, 0 sections   10,604 tokens, 9 sections
```

Flat across a 40x budget range before, responsive after. My numbers differ from yours, the shape does not, which is what I would expect on a different application.

## Containment

I called `readSource` directly with the shapes that usually get through:

| | |
|---|---|
| a file inside the project | read |
| a file outside it | refused |
| `$project/app/../../etc/hosts` | refused |
| a symlink inside the project pointing outside | refused |
| the project root itself | refused |
| `$project-evil/app/In.php`, a sibling directory sharing the prefix | refused |

The sibling case is the one `str_starts_with` normally fails; the `DIRECTORY_SEPARATOR` you appended is doing that work. Both production callers pass `base_path()`, and `ScanCommand` scans `base_path()` too, so the root the export is confined to is the root that was scanned.

## Gate

Output differs on all three applications, and every changed line is the rename. Matching counts on each (182 `source` against 182 `definedIn` on A, 56 against 56 on B, 26 against 26 on C), no node or edge id added or removed, and zero changed lines that are not the rename on any of the three.

For the consumer question: the shipped bundle has no `data.source` in it at all. The other `index-*.js` files in `resources/assets/assets/` are stale builds, which is where a grep finds hits.

`258 passed (859 assertions)`, PHPStan clean, Pint clean, on the head SHA.

## The tests

Reverting `readSource` to return `''` kills the first test. Deleting the containment block kills the third. Both bite.

The other two do not pin what their names say. Delete the `$file === '' || ! is_file($file) || ! is_readable($file)` guard entirely and all four still pass, because a missing path and an empty path both make `realpath()` return `false`, so the containment check refuses them first. The guard's own behaviour is untested. It matters because without it, an exporter constructed with the default empty `projectPath` would reach `file_get_contents()` on a missing file and raise a warning, which under `HandleExceptions` is an `ErrorException`. Constructing one with `projectPath: ''` in those two tests would pin the guard.

## The same file, emitted repeatedly

Nothing keys the emitted set by file, and nodes share files heavily: on app A, 3,491 nodes name 962 distinct files, and one file is named by 441 of them; on app B, 2,782 nodes over 1,125 files, 528 of which more than one node names.

So a selection holding several nodes from one file emits that file's whole contents once per node. On app B's whole-project export, 29 source sections carry 26 distinct bodies, and 59,232 characters (19% of the code emitted) repeat a body already in the export. On a selection concentrated in one file the ceiling is worse than that.

This still beats what it replaces, which emitted no code at all. But the duplicates come out of the budget you just made meaningful, and a `$seen[$file]` check in that loop is a small change. Your call whether it belongs here or in the follow-up you already flagged.

## Cost

Export on app B, interleaved, three reps each: 37–39 ms before, 39–66 ms after; peak memory 20 MB against 22 MB, for an export 4.5× larger. Nothing to worry about.

## Verdict

Approving. The defect is real, the fix reads the right key, the rename is safe and provably confined to itself, and the containment check holds against the cases that usually break one. The two notes above are worth a follow-up commit, not a block.

Also worth saying: the "One thing this does not fix" section is the right way to write one of these. I checked it and it is accurate: the structural section is emitted ahead of the budget on both arms, at any budget.

webard:
Both notes are done in `e2e58a1` — thank you for measuring the duplication rather than describing it, the 19%/26-distinct-bodies number is what made it obvious this belonged here and not in the follow-up.

**Dedupe.** Keyed by the path the node names and checked before the read, so a repeat costs neither budget nor I/O. I kept the key on the raw path rather than the resolved one: two spellings of the same file would still both emit, but that costs a `realpath()` per node to close and the scan writes one spelling.

**The guard.** You were right that neither test reached it, and the fix was smaller than the reason it was needed. Both now construct the exporter unconfined, which is the only way past the containment check.

One thing I had to add beyond your suggestion: `projectPath: ''` alone does not kill the mutant for the *missing-file* half. `file_get_contents()` on a missing path returns `false`, so the export is empty either way and `not->toContain('## Source:')` passes with the guard deleted — it just emits a warning on the way. So the test now installs the handler `HandleExceptions` installs and asserts nothing is raised, which is the actual production difference you named. With that, deleting the guard kills both tests; the empty-path half you suggested kills on its own.

Re-verified on `e2e58a1`: `259 passed (860 assertions)`, PHPStan clean, Pint clean.

SanderMuller (approved):
Re-reviewed at `e2e58a19`. My earlier approval described `08496eb2`, so this replaces it rather than adding to it. I re-ran the probes that produced the two notes, not a fresh reading.

**Duplication, the probe that found it.** Whole-project export at budget 100,000:

| | before | now |
|---|---|---|
| app A | 7 sections, 7 distinct | 14 sections, 14 distinct, 0 duplicated |
| app B | 29 sections, 26 distinct, 59,232 duplicated chars (19% of the code) | 38 sections, 38 distinct, **0 duplicated** |

The budget the copies were consuming now buys more files: distinct files in the same export went from 7 to 14 on A and 26 to 38 on B. Keying on `$file` rather than the node also means a file named by a hundred nodes appears once, which is the case that made this worth doing.

**No over-correction.** A focal route on B still behaves: 2,401 / 2,401 / 6,008 / 10,333 tokens at budgets 500 / 2,000 / 6,000 / 20,000, with 0 / 0 / 1 / 7 source sections and the focal node's file first whenever any is emitted. The 20,000 row shows 7 sections where it showed 9 before, which is the same seven distinct files with two duplicates removed.

**The guard is pinned now.** Deleting `$file === '' || ! is_file($file) || ! is_readable($file)` used to kill nothing. It now fails two tests. Deleting the `$emitted` check fails the new one. Both mutations verified with `git diff --stat` before I trusted the run.

Using `set_error_handler` to turn the warning into what `HandleExceptions` would make of it is a better test than the one I suggested. It pins the reason the guard exists rather than its effect.

**Gate, re-run because a fix commit can move output.** Still rename-only on three applications: 182/182, 56/56, 38/38 `source` against `definedIn`, equal line counts, zero non-rename changed lines. Application C reports 38 where it reported 26 last week because the application itself has grown since, not because anything here changed.

`259 passed (860 assertions)`, PHPStan clean, Pint clean. Approving the current head.

PR #104, opened 2026-08-31
Fix: dependency tables spend the context budget instead of fitting into it

webard:
## The dependency tables spend the budget instead of fitting into it

> Builds on #103, which is what makes the budget mean anything at all. Merge that one first.

Both package tables are built into the "always included" structural block and then subtracted from the budget:

```php
$structural = implode("\n", $parts);   // includes both tables
$charBudget -= strlen($structural);
```

On an application with 120-odd dependencies the tables are long enough to be the entire export. A route asked for at `--budget=500` came back at **1,264 tokens — 2.5× the budget — of which 87% was those two tables and none was the code the route runs.**

They say what the project has installed. They do not say anything about the focal node, which is what the export exists to describe.

### Fitting them

They now take at most 15% of what remains after the structural essentials, and state how many rows they dropped:

```
| …and 96 more | | |
```

Truncated rather than omitted: knowing the project runs Filament at all is worth a few lines, and the count says plainly what was left out instead of implying the list is whole.

Same route, before → after:

| budget | before | after |
|---|---|---|
| 500 | 1,264 tokens · 87% packages · 0 source | **500 tokens · 10% · 1 source** |
| 2,000 | 1,935 tokens · 57% packages · 1 source | 1,884 tokens · 14% · 2 source |
| 6,000 | 5,441 tokens · 20% packages · 4 source | 5,416 tokens · 16% · 4 source |
| 20,000 | 14,174 tokens · 8% packages · 19 source | unchanged — the tables fit whole |

The 500 row is the one that matters: the *tables* now fit inside what the budget leaves, and what is left is spent on code rather than on a dependency list.

Not the export as a whole — that would overclaim. The structural node and edge listing is emitted ahead of the budget and can exceed it on its own (2,923 characters against a 2,000-character limit on one reviewed application, before any table or snippet). This PR does not touch that and does not claim to.

At 20,000 nothing changes, and nothing should — there the tables fit and no truncation row appears.

### Tests

3 tests: at a small budget the focal node's code is still present, the dropped count is stated, and the export stays inside `budget * 4` characters — the same ratio the exporter budgets with.

### Verification

Measured against a real 120-dependency application at four budgets, before and after. `258 → 261 passed`, PHPStan 0, Pint clean.

### DIFF
--- a/src/Ai/ContextExporter.php
+++ b/src/Ai/ContextExporter.php
@@ -17,6 +17,11 @@ class ContextExporter
         'filament_panel', 'filament_resource', 'filament_page', 'filament_widget', 'filament_relation_manager',
     ];
 
+    /**
+     * How much of what is left after the structural essentials the dependency tables may take.
+     */
+    private const PACKAGE_BUDGET_SHARE = 0.15;
+
     public function __construct(
         private readonly GraphStore $store,
         private readonly string $projectPath = '',
@@ -273,35 +278,27 @@ class ContextExporter
             $parts[] = '';
         }
 
-        // Backend packages (composer.json)
-        $composerPackages = $this->readComposerPackages();
-        if (! empty($composerPackages)) {
-            $parts[] = '## Backend Packages (composer.json)';
-            $parts[] = '| Package | Version | Dev |';
-            $parts[] = '|---------|---------|-----|';
-            foreach ($composerPackages as ['name' => $name, 'version' => $version, 'dev' => $dev]) {
-                $devMark = $dev ? 'yes' : '';
-                $parts[] = "| {$name} | {$version} | {$devMark} |";
-            }
-            $parts[] = '';
-        }
-
-        // Frontend packages (package.json)
-        $frontendPackages = $this->readFrontendPackages();
-        if (! empty($frontendPackages)) {
-            $parts[] = '## Frontend Packages (package.json)';
-            $parts[] = '| Package | Version | Dev |';
-            $parts[] = '|---------|---------|-----|';
-            foreach ($frontendPackages as ['name' => $name, 'version' => $version, 'dev' => $dev]) {
-                $devMark = $dev ? 'yes' : '';
-                $parts[] = "| {$name} | {$version} | {$devMark} |";
-            }
-            $parts[] = '';
-        }
-
         // Pre-budget structural content (always included)
         $structural = implode("\n", $parts);
         $charBudget -= strlen($structural);
+
+        // The dependency lists say what the project has installed, not what the focal node
+        // does, and on a large application they are long enough to be the whole export: a
+        // route asked for at 500 tokens came back at 1,264, of which 87% was these two
+        // tables and none was code. They get a share of what is left and no more.
+        $packageBudget = (int) ($charBudget * self::PACKAGE_BUDGET_SHARE);
+        $packages = $this->packageSection(
+            '## Backend Packages (composer.json)',
+            $this->readComposerPackages(),
+            $packageBudget,
+        );
+        $packages .= $this->packageSection(
+            '## Frontend Packages (package.json)',
+            $this->readFrontendPackages(),
+            $packageBudget - strlen($packages),
+        );
+        $structural .= $packages;
+        $charBudget -= strlen($packages);
 
         // Source snippets (budget-gated, focal node first).
         //
@@ -348,6 +345,47 @@ class ContextExporter
         }
 
         return $structural.implode("\n", $sourceParts);
+    }
+
+    /**
+     * One dependency table, as much of it as $charBudget allows.
+     *
+     * Truncated rather than dropped: knowing the project runs Filament at all is worth a few
+     * lines, and the count says plainly what was left out instead of implying the list is whole.
+     *
+     * @param  list<array{name: string, version: string, dev: bool}>  $packages
+     */
+    private function packageSection(string $heading, array $packages, int $charBudget): string
+    {
+        if ($packages === [] || $charBudget <= 0) {
+            return '';
+        }
+
+        $rows = [$heading, '| Package | Version | Dev |', '|---------|---------|-----|'];
+        $used = strlen(implode("\n", $rows));
+        $shown = 0;
+
+        foreach ($packages as ['name' => $name, 'version' => $version, 'dev' => $dev]) {
+            $row = '| '.$name.' | '.$version.' | '.($dev ? 'yes' : '').' |';
+            if ($used + strlen($row) + 1 > $charBudget) {
+                break;
+            }
+            $rows[] = $row;
+            $used += strlen($row) + 1;
+            $shown++;
+        }
+
+        if ($shown === 0) {
+            return '';
+        }
+
+        $omitted = count($packages) - $shown;
+        if ($omitted > 0) {
+            $rows[] = "| …and {$omitted} more | | |";
+        }
+        $rows[] = '';
+
+        return "\n".implode("\n", $rows);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


SanderMuller (changes requested):
Reviewed at `4767264b`. The first commit here is #103, which I reviewed and approved separately; this covers the whole diff, and everything below about behaviour is the delta on top of it.

The core of this is right, and it reproduces on my applications. One thing needs changing before I can approve: a package table can disappear from the export with nothing saying it was ever there, which is the failure mode this PR argues against.

## What I need changed

Give one application a `composer.json` and a `package.json` with 120 dependencies each, and ask for one node:

```
budget    backend rows   frontend rows
   500          5            absent
 1,000         13            absent
 2,000         27            absent
 4,000         57            absent
 8,000        117            absent
20,000        120            120
```

On this branch's parent, both lists come out complete at all six budgets.

The backend table takes the 15% share first, and the frontend call gets `$packageBudget - strlen($packages)`, which by then is nothing. `packageSection()` returns `''`, and an empty string leaves no trace: a reader cannot tell a project with no `package.json` from one whose list did not fit. That is the distinction your own truncation row exists to preserve.

The same path drops the backend table on a real application: at `--budget=500` on app A the structural block alone is 2,923 characters, so `$charBudget` is already negative when the tables are built, `$packageBudget` is negative, and both tables vanish silently. So it is one defect with two instances, not two.

What would change my mind: any mechanism that leaves a trace. Splitting the share between the two tables works, and so does emitting the heading with `| …all 120 omitted | | |` when nothing fits. Either one wants a test on the frontend path, which nothing currently covers: `exporterWithManyPackages()` writes `composer.json` only.

## What I verified, and what holds

The headline result reproduces on app B, one route, both arms, same store:

| budget | before | after |
|---|---|---|
| 500 | 2,401 tokens (4.8x the budget), 94% packages, 0 source | 508 tokens (1.0x), 9% packages, 1 source |
| 2,000 | 2,401 tokens, 94% packages, 0 source | 2,008 tokens, 14%, 1 source |
| 6,000 | 6,008 tokens, 37% packages, 1 source | 6,008 tokens, 15%, 1 source |
| 20,000 | 10,604 tokens, 21% packages, 9 source | unchanged, no truncation row |

Different application, different numbers, same shape as yours, including the 20,000 row where nothing moves and nothing should. Your reading of where the budget went was right, and the fix does what it says at the budgets that matter.

Gate: output differs on all three applications, and the difference is entirely #103's `source` to `definedIn` rename, with matching counts (182/182, 56/56, 26/26), equal dump lengths, and zero non-rename changed lines. This commit adds nothing to graph output, which is what I expected from a change confined to `src/Ai`.

`261 passed (864 assertions)`, PHPStan clean, Pint clean, on the head SHA. CI has not run on this branch, so that is a local result.

Mutations: forcing `$packageBudget` to `PHP_INT_MAX` kills all three new tests, and deleting the `…and N more` row kills the second one alone. Both bite.

## Two notes that are not blocking

**The export is 32 characters over `budget * 4`, always.** Not rounding: the same +32 on both applications at every budget I tried.

```
budget    500   limit  2,000   actual  2,032
budget  2,000   limit  8,000   actual  8,032
budget  6,000   limit 24,000   actual 24,032
```

It is the truncation marker. `$available = $charBudget - $overhead` sizes the snippet to fill the budget exactly, and then `"\n// [truncated — token budget]"` is appended after the arithmetic. That marker is 32 bytes, since the em dash costs three. Subtracting `strlen()` of that marker from `$available` brings it to +2.

This predates your branch, but your third test now names it as an invariant, and it passes only because the fixture file is small enough never to truncate. Add 500 filler lines to `Demo.php` and it fails at 2,032 against 2,000, which is where I got the number.

**"The export now fits the budget it was given" is true of the tables, not of the export.** On app A at `--budget=500` the node and edge listing alone is 2,923 characters, 46% past the limit, before any table or snippet. Your tables now fit inside what is left; the remaining offender is the structural block, which this PR does not touch and does not claim to. Worth narrowing the sentence, not worth doing anything about here.

webard:
Fixed in `c66f2dd`. You were right, and it is worth saying plainly: the defect you found is the one this PR is about. I moved the tables off the critical path and then let one of them fall off the edge of it.

**Split, not first-come.** Each table now gets half the share, and whatever the first leaves unspent rolls forward — so a project with only a `composer.json` still gets all 15%, and one with both gets a table each.

**A table that does not fit at all now leaves its heading and its count.** I took your second suggestion as well as the first, because the split alone does not cover the case you found second: when the structural block has already overrun the budget the share is negative and *both* tables vanish. You called it one defect with two instances, and that is where the trace belongs — at the bottom, in `packageSection()`, not in the caller.

**The +32.** Fixed, and thank you for pinning it to the marker rather than to rounding — that saved me looking in the wrong place. `$available` now accounts for the marker before `substr()`. Your point about my own test is the sharper one: it named an invariant it could not have caught failing, because the fixture never truncated. It now writes 500 filler lines, and I confirmed it fails at 2,032 against 2,000 with the fix reverted.

**On the frontend test.** Its first version asserted the heading and passed with the share unsplit — the omission trace prints the heading too, so the two changes covered for each other. It asserts package rows now, and the unsplit mutant dies.

**And the sentence.** Narrowed: the tables fit the budget, the export does not. The structural block is 46% past the limit on your app A before anything else is added, and this PR neither touches that nor claims to.

Re-verified on `33e9bf9`: `264 passed (869 assertions)`, PHPStan clean, Pint clean. The branch also now carries #103's follow-up commit, since it sits on top of it.

SanderMuller (approved):
Re-reviewed at `33e9bf9a`. I re-ran the probes that produced the findings rather than reading the diff for them. The blocking one is fixed, and so are both notes.

**The vanishing table.** Same fixture: 120 composer dependencies, 120 npm, one node.

| budget | before | now |
|---|---|---|
| 500 | backend 5 rows, frontend **absent** | backend 2 rows + count, frontend heading + "all 120 omitted" |
| 1,000 | backend 13, frontend absent | backend 5 + count, frontend 6 + count |
| 2,000 | backend 27, frontend absent | 13 and 12 |
| 8,000 | backend 117, frontend absent | 58 and 57 |
| 20,000 | 120 and 120 | 120 and 120, no truncation row |

The split is even, the leftover rolls forward so a project with only one file still gets the whole share, and the case where nothing fits leaves a heading and a count instead of nothing. The second instance I reported behaves too: on app A at `--budget=500`, where the structural block alone is already past the limit, both tables now leave a trace where both used to disappear.

**The 32 characters.** Gone, and not approximately:

```
budget    500   limit  2,000   actual  2,000   over 0
budget    800   limit  3,200   actual  3,200   over 0
budget  1,200   limit  4,800   actual  4,800   over 0
budget  2,000   limit  8,000   actual  8,000   over 0
budget  6,000   limit 24,000   actual 24,000   over 0
```

Both applications, every budget, exactly on the line. Subtracting the marker inside the arithmetic was the right place for it.

**The tests bite, each on its own axis.** Reverting `$firstShare` to the whole share kills the frontend test alone. Returning `''` for a table that does not fit kills the "names a table it could not fit" test alone. Appending the marker after the arithmetic again kills "keeps the export inside the budget", the test that used to pass whatever the code did, because the fixture never truncated. It does now. Every mutation confirmed with `git diff --stat` before I believed the run.

**What is left, and it is yours to decide.** Across 360 routes at `--budget=500` on three applications, 20 exports still exceed `budget * 4`, the worst by 1,767 characters. That one is 3,767 characters of which 3,610 is the node and edge listing, 157 is the two table traces, and none is source. So the tables are behaving and the structural block is now the sole offender, which is what your own "One thing this does not fix" says. Nothing here made it worse.

Gate re-run because a fix commit can move output: still rename-only on three applications, 182/182, 56/56, 38/38, zero non-rename changed lines. `264 passed (869 assertions)`, PHPStan clean, Pint clean.

Approving. This one covers the whole stack, #103 included, which I re-approved at its own head.

PR #105, opened 2026-08-31
Fix: say when --budget is raised to the minimum

webard:
## `--budget` is silently raised to a minimum the help does not mention

`brain:export-context` clamps the option:

```php
budget: max(500, (int) $this->option('budget')),
```

and the help reads `{--budget=6000 : Token budget}` — no minimum anywhere. So `--budget=300` becomes 500 without a word, and the export's own header then reports `Budget: 500 tokens`, a number the caller never chose. Someone tuning the budget down has no way to know their value stopped taking effect.

The minimum is in the help now, and the command says so when it actually applies:

```
WARN  Budget raised to the 500-token minimum (asked for 300).
```

Nothing is printed when the requested budget is used as given, so the normal case is unchanged.

The clamp itself stays: below 500 tokens the export cannot carry a focal node and its immediate context, which is the least it can be useful at. The complaint was never the floor, only that it was invisible.

### Verification

Run against a real project: `--budget=300` warns and exports at 500, `--budget=6000` exports with no warning. `254 passed`, PHPStan 0, Pint clean.

No test: this is a help string and one console line, and the console output is the behaviour. Happy to add one if you would rather have it pinned.

### DIFF
--- a/src/Commands/DiagnosticStream.php
+++ b/src/Commands/DiagnosticStream.php
@@ -1,0 +1,34 @@
+<?php
+
+declare(strict_types=1);
+
+namespace LaraMint\LaravelBrain\Commands;
+
+use Symfony\Component\Console\Output\ConsoleOutputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+/**
+ * The stream a diagnostic belongs on when the command's real output is an artifact.
+ *
+ * `brain:export-context` writes a document to stdout that is meant to be redirected into a file
+ * or piped into something. Anything else written there lands INSIDE that document: a warning
+ * became the first line of the export, ahead of its own heading, and with `--format=json` it made
+ * the result unparseable. Diagnostics go to stderr so that stdout carries the artifact alone.
+ *
+ * The `instanceof` check is the load-bearing part. `OutputStyle::getOutput()` returns whatever the
+ * command was given, and only a `ConsoleOutputInterface` has a separate error stream — a buffered
+ * output, which is what Laravel hands a command under test, does not, and calling
+ * `getErrorOutput()` on one is a fatal error rather than a graceful miss.
+ *
+ * `OutputStyle::getErrorOutput()` exists but is `protected`, which is why this asks the underlying
+ * output rather than the style wrapping it.
+ */
+final class DiagnosticStream
+{
+    public static function for(OutputInterface $output): OutputInterface
+    {
+        return $output instanceof ConsoleOutputInterface
+            ? $output->getErrorOutput()
+            : $output;
+    }
+}
--- a/src/Commands/ExportContextCommand.php
+++ b/src/Commands/ExportContextCommand.php
@@ -40,15 +40,22 @@ class ExportContextCommand extends Command
         $exporter = new ContextExporter($store, base_path());
 
         try {
-            $requestedBudget = (int) $this->option('budget');
+            $rawBudget = (string) $this->option('budget');
+            $requestedBudget = (int) $rawBudget;
             $budget = max(self::MINIMUM_BUDGET, $requestedBudget);
 
             // Silently exporting more than was asked for is worse than saying so: the header
             // of the export itself reports the budget, and it would report a number the
             // caller never chose.
+            //
+            // On stderr, not stdout — stdout is the export. See {@see DiagnosticStream}.
+            // And quoting back what was typed rather than what it cast to: `--budget=abc` is 0
+            // as an int, and "asked for 0" names a number nobody wrote.
             if ($budget !== $requestedBudget) {
-                $this->components->warn(
-                    "Budget raised to the {$budget}-token minimum (asked for {$requestedBudget})."
+                $asked = is_numeric($rawBudget) ? $rawBudget : '"'.$rawBudget.'"';
+
+                DiagnosticStream::for($this->output->getOutput())->writeln(
+                    "<comment>Budget raised to the {$budget}-token minimum (asked for {$asked}).</comment>"
                 );
             }
 
--- a/src/Commands/ExportContextCommand.php
+++ b/src/Commands/ExportContextCommand.php
@@ -10,10 +10,13 @@ use LaraMint\LaravelBrain\Storage\GraphStoreFactory;
 
 class ExportContextCommand extends Command
 {
+    /** Below this the export cannot carry a focal node and its immediate context. */
+    private const MINIMUM_BUDGET = 500;
+
     protected $signature = 'brain:export-context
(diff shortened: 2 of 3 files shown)


SanderMuller (changes requested):
Reviewed at `50f8c608`. The diagnosis is right and the clamp boundary behaves exactly as described. One thing blocks it: the warning goes to stdout, which is where the export goes.

## The warning lands inside the artifact

I booted the command for real against a small store and separated the streams:

```
$ php artisan brain:export-context --budget=300 > context.md 2> err.txt
$ head -3 context.md

   WARN  Budget raised to the 500-token minimum (asked for 300).

$ wc -c err.txt
0 err.txt
```

`$this->components->warn()` renders through `$this->output`, the same `OutputStyle` that `$this->line($output)` writes the export to a few lines later. So the banner is the first thing in the file, ahead of `# Laravel Brain AI Context`, in a document whose entire purpose is to be handed to a model. stderr stays empty.

With `--format=json` it is worse, because the result is no longer parseable:

```
$ php artisan brain:export-context --budget=300 --format=json | jq .
parse error: Invalid numeric literal at line 1, column 8
```

At `--budget=6000` the same command gives valid JSON. The README shows bare invocations for both formats (lines 121, 124, 127, 133), so piping is a documented path, not an exotic one. `--output=` is unaffected: that file is clean, and the warning belongs on the console there.

What would change my mind: send it to stderr. I tried the obvious `$this->output->getErrorOutput()`, which does not compile, because Symfony's `OutputStyle::getErrorOutput()` is protected. This works, and I ran it:

```php
$stderr = $this->output->getOutput();
if ($stderr instanceof \Symfony\Component\Console\Output\ConsoleOutputInterface) {
    $stderr = $stderr->getErrorOutput();
}
$stderr->writeln("<comment>Budget raised to the {$budget}-token minimum (asked for {$requestedBudget}).</comment>");
```

With that in place, stdout is the export alone and stderr carries the line. The `instanceof` guard matters because a buffered output in a test is not a `ConsoleOutputInterface`, and calling `getErrorOutput()` on one would fatal.

## What I checked and found fine

Every boundary of the clamp, run end to end:

| invocation | warns | header |
|---|---|---|
| `--budget=6000` | no | Budget: 6000 tokens |
| `--budget=500` | no | Budget: 500 tokens |
| `--budget=499` | yes, "asked for 499" | Budget: 500 tokens |
| `--budget=-100` | yes, "asked for -100" | Budget: 500 tokens |
| no `--budget` at all | no | Budget: 6000 tokens |

The off-by-one is right at both ends, the default path is untouched, and `--output` writes a file with no banner in it while the console still says what happened. The help string now carries the minimum, and the constant is named where it is used.

Gates: this touches `src/Commands` only, so the output gate has nothing to see and I did not run it. `254 passed (853 assertions)`, PHPStan clean, Pint clean, which is your number exactly. CI has not run on this branch, so that is a local result. The branch is level with `v2.x`, no stale base.

On your offer of a test: I would not hold the PR for one. The behaviour that needs pinning is which stream a line goes to, and Laravel's in-process command testing buffers both into one string, so a test would have to spawn a process to say anything real. The `--output` path is easy to assert on if you want something.

## One small thing

`--budget=abc` casts to 0 and reports `asked for 0`, a number the caller never typed. Reading the raw option and saying `asked for "abc"` when it is not numeric would be more honest, and it is the same line either way.

webard:
Fixed in `dfd38e6`. You are right and it is the worse half of the bug I was fixing: I wrote a line to say the budget was not what the caller asked for, and put it in the document whose header reports the budget.

Your snippet, with the guard, moved into a named class — `DiagnosticStream` — because the reason it needs the guard is worth keeping next to it, and the caller reads better for it. Reproduced your two commands on a real application:

```
$ brain:export-context --budget=300 > context.md 2> err.txt
before   context.md line 1: WARN  Budget raised to…        err.txt   0 bytes
after    context.md line 1: # Laravel Brain AI Context     err.txt  56 bytes

$ brain:export-context --budget=300 --format=json | jq .
before   parse error: Invalid numeric literal at line 1, column 8
after    parses; keys meta, packages, nodes, edges
```

`--budget=6000` writes nothing to stderr, and the `--output=` file is unchanged — the warning still reaches the console there, which is where you said it belongs.

**`--budget=abc`** now quotes back what was typed: `asked for "abc"`. Same line, one `is_numeric()`.

**On the test.** You were right that the streams cannot be separated from a command test, and I did not add a harness to try — `tests/Feature` is empty and the package has nothing Laravel-booting in it, so a `--output` assertion would have meant standing one up for a single line. What can be pinned without a process is the choice itself, so there are two tests on `DiagnosticStream` instead. Both bite: dropping the redirect kills the first, and dropping the guard kills the second with exactly the fatal it exists to prevent (`Call to undefined method BufferedOutput::getErrorOutput()`) — which is also the clearest statement I could make of why the guard is there.

Re-verified on `dfd38e6`: `256 passed (856 assertions)`, PHPStan clean, Pint clean.

SanderMuller (approved):
Re-reviewed at `dfd38e69`. I re-ran the probe that produced the finding: the real command, booted against a small store, with the two streams redirected to separate files.

```
$ php artisan brain:export-context --budget=300 > context.md 2> err.txt
$ head -1 context.md
# Laravel Brain AI Context
$ cat err.txt
Budget raised to the 500-token minimum (asked for 300).
```

The banner is off the artifact. The JSON case parses again: `--budget=300 --format=json | jq .` succeeds, with the warning on stderr where `jq` never sees it. `--output=` still writes a clean file and still tells the console what happened, now on stderr, which is right for a run whose stdout says only "Context written to …".

Boundaries unchanged, all six run end to end: silent at 500, at 6,000 and with no `--budget` at all; warns at 499 and at -100; and `--budget=abc` now reads `asked for "abc"` rather than naming a 0 nobody typed.

`DiagnosticStream` is a better answer than the inline version I suggested, because the `instanceof` is the whole point and a named class is where that reasoning survives. Both branches bite: returning `$output` unchanged fails the console test, and dropping the guard to call `getErrorOutput()` unconditionally fails the buffered test with exactly the fatal it exists to prevent (`Call to undefined method BufferedOutput::getErrorOutput()`). Both mutations confirmed with `git diff --stat` before I read the run.

One thing still true and worth saying out loud: no test asserts that the *command's* stdout is clean, because Laravel's in-process command testing buffers both streams into one string. The unit tests pin the routing decision, which is the part that can regress silently; the end-to-end behaviour is what I ran by hand above. That is the right split, not a gap to fill.

`256 passed (856 assertions)`, PHPStan clean, Pint clean. Output gate still not applicable, `src/Commands` only. Approving.

PR #106, opened 2026-08-31
Chore: drop 398 build assets nothing references (69.5 MB)

webard:
## `resources/assets/assets/` carries every bundle ever built

The viewer loads seven files. The directory tracks 405.

| | files | size |
|---|---:|---:|
| referenced | 7 | 0.8 MB |
| orphaned | **398** | **69.5 MB** |

345 stale `index-*` chunks, 40 `html2canvas-*` from a build that split it into its own chunk, 12 `vendor-*`, one `rolldown-runtime-*`. Vite emits a new content-hashed name on every build and the old one is never removed, so each commit that rebuilt the frontend added a full set and kept the previous ones. Every `composer require` of this package downloads all 70 MB.

### Establishing what is actually reachable

Reading the blade alone is not enough — a chunk can pull another in by name, and `html2canvas` is a dynamic import (`await import('html2canvas')` in `exportUtils.ts`), so its chunk could be reachable without ever appearing in the HTML.

So the live set was computed as a transitive closure: seed from `index.blade.php`, then follow any tracked asset filename appearing inside a live chunk. It converges on the same seven. The current build folds html2canvas into `vendor-utils`, and the only `import(...)` in the live entry chunk names that file:

```js
import(`./vendor-utils-D7YtnGoz.js`
```

The 40 standalone `html2canvas-*.js` files are leftovers from a build that chunked it differently.

### Verified in a browser, not only by reading

With the directory pruned to the seven and the package served to a real application: the viewer renders, every `script[src]` request returns 200, there are no console errors, and the lazily imported `vendor-utils` chunk resolves — which is the one that would have broken had the closure been wrong.

`254 passed`, PHPStan 0, Pint clean — unchanged, as nothing in `src/` is touched.

### Worth deciding alongside this

Pruning once fixes the current state, not the mechanism: the next frontend PR adds a new hashed set and keeps the old one. Either the build step should remove what the manifest no longer names, or the directory should be regenerated rather than accumulated. I have not touched `npm run build` here because that is a workflow decision rather than a cleanup, and it belongs to whoever maintains the release process — happy to follow up once you say which shape you want.

Git keeps the history, so nothing here is unrecoverable.


SanderMuller (approved):
Reviewed at `db7a8ae5`. For a 398-file deletion I did not want to check your closure, I wanted to compute my own and see whether it agreed. It does, exactly.

## Recomputed from scratch

Seeded from every filename mention of a tracked asset anywhere in `resources/views`, `src`, `routes`, `config`, `tests`, `docs` and the README, then followed any tracked asset filename appearing as a substring inside a live file until the set stopped growing. Substring matching over-approximates deliberately: a file gets kept if its name appears anywhere at all, so the closure errs toward keeping things.

It converges on 8: your 7, plus `logo.png`, which the blade uses as the favicon and the entry chunk also names.

Set difference against your diff:

```
tracked assets                            408
in my closure                               8
deleted by this PR                        398
deleted but inside my closure               0
in neither (kept anyway)   favicon.png, icons.svg
```

Zero overlap between what you deleted and what I compute as reachable. The two files we differ on you kept, so the difference is on the safe side.

Your load-bearing point was the dynamic import, and it holds. The live entry chunk contains exactly one `import(`:

```
import(`./vendor-utils-D7YtnGoz.js`
```

which is kept. The 40 `html2canvas-*` chunks are only ever named by *other* stale `index-*` chunks, which is why a grep across the whole directory looks alarming and means nothing. The stylesheet contains no `url(...)` at all, so no font or image hangs off it.

## Served, not just reachable

I resolved each URL the blade requests through `BrainController::packageAssetPath()` on the pruned tree, which is the same path resolution the `/{any?}` route uses:

```
assets/index-Bze1eHA4.js              RESOLVES   131,640 bytes
assets/rolldown-runtime-BHe-jwch.js   RESOLVES       957
assets/vendor-CsjAK7B8.js             RESOLVES   135,169
assets/vendor-react-CNktMmef.js       RESOLVES   221,549
assets/vendor-d3-DThTr3c3.js          RESOLVES    57,917
assets/vendor-utils-D7YtnGoz.js       RESOLVES   199,570
assets/index-CHLc6t41.css             RESOLVES    71,041
logo.png                              RESOLVES   419,642
```

8 of 8. Nothing publishes these to an application's `public/`, so every install serves them from the package directory and there is no second copy to go stale.

I also checked the upgrade path, since a deletion can strand a cached view: Blade's `isExpired()` compares the source mtime against the compiled one, so an application that upgrades recompiles `index.blade.php` and asks for the new names. No `view:clear` needed.

## Numbers

All three reproduce on the merge-base tree:

| | files | size |
|---|---:|---:|
| referenced | 7 | 0.8 MB |
| orphaned | 398 | 69.5 MB |
| total | 405 | 70.3 MB |

There is no `.gitattributes`, so nothing was export-ignoring this directory already and the install really does carry all 70 MB today.

`254 passed (853 assertions)`, PHPStan clean, Pint clean. Nothing outside `resources/assets/` is touched, which I checked rather than assumed. The output gate does not apply here and I did not run it: no analyzer, parser or graph code changes, so a byte-identical result would say nothing.

## Approving

No blocking finding. Two small things, neither worth holding this:

`favicon.png` and `icons.svg` are unreferenced by the same scan that cleared the other 398: the blade points `rel="icon"` at `logo.png`. Keeping them is the conservative call and I would keep them too, but they are probably dead as well.

On the mechanism: I agree with your read, and it is the right call to leave it out of a cleanup PR. Worth saying that whatever shape it takes, the thing to prevent is precisely this review, where the only way to know which seven of 398 files matter was to build a closure.

webard:
Approved with two notes, and I am leaving both as they are — but I recomputed rather than taking either on trust.

**`favicon.png` and `icons.svg`.** Your scan is right: zero references, and I confirmed it holds in `frontend/` too, which is the build SOURCE and the one place a reference could hide from a scan of the shipped tree. `logo.png` is named in five places (`frontend/index.html`, `App.tsx`, `Toolbar.tsx`, the blade's `rel="icon"`, and the live entry chunk); the other two in none.

I am still keeping them, and not only out of caution. They are hand-authored source assets sitting beside `logo.png`, not build output — this PR deletes files Vite emitted and abandoned, which is a category with a mechanical test for deadness. "Nobody currently references this icon" is a different claim, it needs a maintainer's intent rather than a closure, and 13.5 KB does not buy the right to answer it inside a 69.5 MB cleanup. If they should go, they should go in a PR that says so.

**On the mechanism.** Your closing line got a data point sooner than either of us expected. My integration branch, which carries this PR plus the four layout fixes, had regrown four orphans by the time I looked — one per `npm run build`, since each emits a new content-hashed entry chunk and nothing removes the one it replaces:

```
v2.x                      259 index-*.js
+ breadth-first hang      260
+ layer spacing           261
+ unreached components    262
+ circle radius           263
```

Four PRs, four new files nothing will ever reference. I have pruned them on my branch — closure recomputed the way you did it, and I loaded the viewer against a real application to check they are unreachable rather than merely unreferenced: ten requests, all 200, no 404, no console error, and the eight the page asks for are exactly the closure.

I am not folding that into this PR; it belongs with the branches that created them. But it does turn "worth preventing" into "prevents itself in under a week", which is an argument for the mechanism rather than against it.

PR #107, opened 2026-08-31
Fix: circle layout shrinks the space per card as cards are added

webard:
## Before and after

The same 20-node tab in the Circle layout, fit to view, at `b7148f7` and `77a637a`. Playwright-driven, so the pair is one run rather than two sittings.

| before | after |
|---|---|
|  |  |

The before is the `min(280, 90 + n * 4)` radius: twenty cards 185–270px wide placed 53px apart, so the ring is a stack and most of the labels are behind another card.

## The circle layout shrinks the space per card as cards are added

> Builds on #101, and through it #99 and #100. Stacked so the diff stays to the circle; merge those first.

`layoutCircle` places nodes at uniform angles, so each one gets `2πr / n` of arc. The caller picks the radius like this:

```ts
const r = Math.min(280, 90 + nodesCopy.length * 4)
layoutCircle(nodesCopy, r)
```

A ceiling of 280. So the arc available to each card **shrinks as cards are added** — which is backwards, and past about 48 nodes the ceiling binds and every further node makes it strictly worse.

Twenty cards at radius 170 get **53px of arc each**, for cards 185–270px wide. The ring is a pile in the middle of the canvas with the labels unreadable.

### Sizing it from the cards

Every card gets the same arc, so the arc is sized for the largest — and by its **longer side**, since which side faces the tangent depends on where on the ring a card sits:

```ts
const arc = Math.max(...nodes.map((node) => Math.max(node.width, node.height))) + gap
const radius = Math.max(arc, (n * arc) / (2 * Math.PI))
```

The `Math.max(arc, …)` floor keeps a two-node graph from collapsing onto a radius of nearly zero.

Twenty cards go from a radius of 170 to 987, and from 53px of arc each to 310px.

Measured across 576 stored graphs, laying each out both ways with the renderer's own width formula:

| | tabs with overlapping cards |
|---|---:|
| before | **536** |
| after | **0** |

Confirmed in a browser against a real application: the ring is a ring, and the labels are legible.

### The same defect as #100, one function over

#100 fixed the layered layouts and the grid, where a constant pitch was chosen without reference to card size. This is the same mistake expressed as a constant radius. I did not catch it there because I was checking pitch-versus-width and the circle has no pitch — the relationship is `2πr / n`, and the ceiling made it worse with scale rather than merely wrong at one scale.

`layoutDagre` and `layoutForce` remain correct: dagre is given real dimensions, and the force simulation separates by a collision radius derived from them.

### Verification

- Geometric measurement over 576 graphs, before and after.
- Rendered in a browser, no console errors.

The caller stops passing a radius and passes nothing; the signature's second parameter is a gap now, with one call site, as in #100.


SanderMuller (approved):
Reviewed at `77a637af`. This covers the whole diff, which is the four-commit stack: #99, #100, #101 and the circle fix. I reviewed the first three separately; what follows is the combined state, measured rather than re-read.

## The circle claim, recomputed independently

I ran your `layoutCircle` over the stored subgraphs of three applications, both arms, using the repository's own `buildLayoutNode` for the card sizes and the old caller's `Math.min(280, 90 + n * 4)` for the before arm. Overlap is an axis-aligned box test between every pair.

| application | tabs | tabs with overlapping cards, before | after |
|---|---:|---:|---:|
| A | 537 | 520 | **0** |
| B | 367 | 350 | **0** |
| C | 423 | 411 | **0** |

1,327 tabs, 1,281 of them broken before, none after. Different corpus from your 576, same result.

The grid, from #100, holds in the combined state too: 409 of C's 423 tabs overlapped before, none after.

## The hang is real, and smaller than a real graph

The stack's first commit is the one I would have wanted evidence for most, so here it is. Three nodes, `root → a → b → a`:

```
head:  returned in 0.17ms
base:  still running after 2 minutes, killed
```

Laying out one application's 367 tabs takes 5 ms total on this branch, worst tab 0.2 ms. The same run against the parent does not finish.

One precision note on that commit message: it takes a cycle *reachable from a root*, not any cycle. A bare inverse pair on its own gives both nodes an in-degree of 1, so `roots` is empty, the queue never runs, and the old code returns immediately, with the two cards overlapping, which is the other bug rather than this one. Everything real has a controller pointing into the models, so the distinction is academic for users and only matters if someone reads the message as a spec.

## Edges

`layoutCircle` at each size, worst clearance between any two cards (negative would be an overlap):

```
n=0      handled by the early return
n=1      radius 282
n=2      radius 282, clearance 473
n=3      radius 282, clearance 246
n=20     radius 921, clearance 35
n=177    radius 8361, clearance 20
n=1000   radius 47237, clearance 20
```

The `Math.max(arc, …)` floor does what you say it does at two and three nodes, and clearance never goes negative. It settles near 20px rather than the 40px gap because neighbours are separated along a chord while cards are axis-aligned, which is a fair reading of "gap" and not a defect.

`Math.max(...nodes.map(…))` throws `RangeError` past about 124,000 nodes. The largest tab in three applications is 177 nodes, and the largest whole graph is around 10,000, so nothing reachable. Worth knowing it exists rather than discovering it.

## The shipped bundle matches the source

Checked rather than assumed, since a stale chunk would make the whole diff decorative. The referenced chunk carries the new formula:

```js
function W(e,t=40){let n=e.length;if(!n)return;
  let r=Math.max(...e.map(e=>Math.max(e.width,e.height)))+t,
      i=Math.max(r,n*r/(2*Math.PI));
```

and `Math.min(280` appears nowhere in it.

`tsc -b` clean, `npm run lint` clean, `254 passed (853 assertions)`, PHPStan clean, Pint clean. The output gate does not apply and I did not run it: no analyzer, parser or graph code is touched, so a byte-identical result would say nothing.

## Approving, with one thing worth doing before merge

The build ships four `index-*.js` chunks, one per commit, and the blade references one:

```
index-DSQtxLY2.js   132K   not referenced
index-ejFbAAeZ.js   132K   referenced
index-odlQ6dFl.js   132K   not referenced
index-pWcB9eEi.js   132K   not referenced
```

396K of chunks nobody loads, which is the accumulation #106 is removing 398 files of. Rebuilding once on the final commit rather than on each would add one chunk instead of four. Worth noting for whoever merges: after both land, `index-Bze1eHA4.js` also becomes unreferenced and is not in #106's list, since that list was computed while the blade still pointed at it.

Neither is a reason to hold this. The layout defect is real, the fix is the right shape, and it is the first of these four where I could measure the before and after myself and get your answer.

webard:
Done in `80ad4c2`, and thank you for the `index-Bze1eHA4.js` catch — that one I would not have found. It is referenced on `v2.x`, so #106 is right to keep it and its list is not wrong; it is this stack that strands it. Removing it here rather than leaving a sequel to #106, and it is a clean delete in either merge order since #106 does not touch it. Four chunks gone, one entry chunk ships.

Verified the same way you verified the deletion in #106: closure recomputed from the blade outward, landing on the seven that remain with none of the four in it, then the viewer loaded against a real application with this tree in place — ten requests, all 200, no 404, no console error. I also switched a 20-node tab to Circle to confirm the shipped chunk is still the fixed one: cards on a ring, none touching.

**On the cycle wording.** You are right and the distinction is not only academic to me — the commit message says "any graph with a cycle" and the loop needs a cycle *reachable from a root*, because a bare inverse pair leaves `roots` empty and the old code returns immediately with the two cards overlapping. That is #100's bug, not #99's. The message is on a branch you have already reviewed and I would rather not force-push under you; I have corrected the claim in #99's description, where it can be fixed without rewriting anything.

**On the `RangeError`.** Real, and I am deliberately not fixing it in this PR. A `reduce` instead of the spread is one line, but it changes the bundle — including the minified formula you quoted back to prove the shipped chunk matches source — and invalidating that verification to move a cliff that sits 12x beyond the largest graph either of us has measured is a bad trade for an approved PR. I will fold it into whichever change next rebuilds the frontend, and it will be visible in that diff rather than smuggled into a deletion.

Two small corrections to what I claimed, since you measured them: the ~20px settled clearance is a chord-versus-axis-aligned artefact and my "40px gap" reads as the arc separation, not the box separation. Your reading is the accurate one.

SanderMuller:

Twenty or so cards in `circle` on each side. Your description already has the numbers (53px of arc for a 185-270px card, a radius going from 170 to 987), and a picture of the ring as a pile in the middle of the canvas is what those numbers stand for.

SanderMuller (approved):
Re-reviewed at `80ad4c20`. My earlier approval described `77a637af`, so this replaces it.

The delta is one commit and it is the answer to the note I left: four entry chunks deleted, three of them the intermediate builds, the fourth `index-Bze1eHA4.js`, which #107 superseded when the blade moved off it. Nothing else changed.

Checked at the new head:

- The branch now adds exactly one entry chunk against its merge-base, `index-ejFbAAeZ.js`, and the blade points at it.
- `npm ci && npm run build` leaves `git status` empty, so the committed chunk is byte-for-byte what the source produces.
- The circle measurement that produced my approval is unchanged: 0 tabs with overlapping cards out of 537, 367 and 423 on three applications.
- CI is green with per-job results, 13 PHP jobs plus `frontend`.

Taking `index-Bze1eHA4.js` with the others also closes the interaction I flagged with #106: that file was in #106's keep list because the blade still pointed at it when #106 was written, and it would otherwise have survived as an orphan whichever order the two merged in. Approving the current head.

webard:
Added to the description — the 20-node tab in Circle, fit to view, at `b7148f7` and `77a637a`, captured in one Playwright run.

The before is the `min(280, 90 + n * 4)` radius doing what the description says: twenty cards 185–270px wide placed 53px apart, so the ring is a stack and most labels are behind another card. Your overlap table said the same thing for 1,281 tabs; this says it for one, to somebody without a checkout.

PR #108, opened 2026-08-31
Fix: the sidebar counts ignore the search box and the method chips

webard:
> **Stacked on #107**, which is stacked on #99, #100 and #101. The last commit is this PR; the five below it are that stack. No other merged branch touches `frontend/src`, so the stack is frontend-isolated and this diff is the whole of it.

## Filtering the sidebar changed the route tree and nothing else

`Routes`, `Risks` and `Recent` kept their unfiltered numbers. The `Risks` and `Recent` **lists** kept their unfiltered contents too — type a query, switch to `Risks`, and every flagged route in the application is still sitting there.

Measured on a 621-route application:

| | before | after |
|---|---|---|
| chips narrowed to `DELETE` | `Routes 621 · Risks 2 · Recent 0` above a tree of 125 | `125 · 0 · 0` |
| search `logout` | `Routes 621 · Risks 2 · Recent 0` above a tree of one, two cards under `Risks` | `1 · 1 · 0`, one card: `POST /gui/logout` |

The `Routes` count was the honest one only by accident: it read `tabs.length`, and `tabs` is the unfiltered set, so it agreed with the tree exactly when no filter was on.

## Before and after

Captured with Playwright against a 621-route application, driving the sidebar rather than posing it, so the pair comes out of one repeatable run. Same scan data, same viewport, only the commit differs.

**Search `logout`, `Routes` tab.** The count is the application's whole route list; the tree under it has one entry.

| before | after |
|---|---|
|  |  |

**Search `logout`, `Risks` tab.** The second bug, and the one the table above cannot show: the list itself ignored the query, so `/broadcasting/auth` is sitting there under a search for `logout`.

| before | after |
|---|---|
|  |  |

**Chips narrowed to `DELETE`.** `Routes 621` above a tree of 125, and `Risks 2` above what is now an empty set.

| before | after |
|---|---|
|  |  |

The 125 is not 125 `DELETE` routes: console commands, broadcast channels, schedules and Filament pages have no HTTP method, so the method chips do not apply to them and they stay. That is existing behaviour of the tree filter, now applied consistently to the counts as well.

## The cause

Three derivations, three different inputs:

```ts
const tree      = useMemo(() => buildTree(tabs.filter(/* query + methods */)), …)
const riskTabs  = useMemo(() => tabs.filter(…),  [tabs])   // unfiltered
const recentTabs= useMemo(() => tabs.filter(…),  [tabs])   // unfiltered

{ id: 'routes', count: tabs.length }        // unfiltered
{ id: 'risks',  count: highRiskCount }      // a prop from the parent
{ id: 'recent', count: recentCount }        // a prop from the parent
```

The filter now runs once, into `filteredTabs`, and every list and every count comes from it. Each count is taken from the list it labels rather than computed alongside it.

## A second disagreement, which the same change settles

`highRiskCount` is the parent's `high`-and-`critical` tally. The list under that badge shows **everything with a risk level** — `riskOf(t) !== 'none'`. So the badge and its own list disagreed whenever a route was `medium` or `low`, silently, and agreed only on applications where every finding happened to be high. That is why this has been sitting there: on the application I found it on, both findings are `high`, so `2` was right for the wrong reason.

The list is the intentional one — the empty state says "No flagged routes", `FlagCard` renders a severity chip per card, and `RISK_ORDER` exists to sort mixed severities. So the count was simply reading the wrong variable.

**The toolbar's `High-risk` pill is untouched.** It says high-risk, it means high-risk, and it is deliberately global — it is the one number that should not move when you filter.

`highRiskCount` and `recentCount` are gone from this component's props, and `recentCount` from the parent, which had no other consumer. Leaving them would have left the second source of truth that caused this.

## Empty states no longer deny what a filter is hiding

`No flagged routes. ✓` claims a clean bill of health. With a filter on it meant "none matching", which is a different statement and the checkmark makes it the wrong one. Both that and `Nothing changed since the previous scan.` now say which they mean.

## Folded in from #107's review: `Math.max(...array)`

That review noted the spread throws `RangeError` past roughly 124,000 elements, and I said I would fold it into whichever change next rebuilt the frontend. This is it, and it is visible here rather than smuggled into a deletion.

Five call sites across the layered, circle and grid layouts now `reduce` instead. Nothing observed is close to the limit — the largest single tab measured is 177 nodes and the largest whole graph around 10,000 — so this fixes nothing anyone has hit. It is one line either way and a layout that throws is a blank screen. Empty input still yields `-Infinity` exactly as the spread did, and every caller already guards for it, so the behaviour is identical everywhere it is reachable.

## Verified

In a browser, against a real application's scan data, on the built bundle:

- both filters recalculate all three counts, together and separately
- the `Risks` list narrows with its count — `logout` leaves exactly `POST /gui/logout`
- the new empty state appears under a filter, the old one still appears without
- Circle still lays out with no overlapping cards after the `reduce` change
- console clean throughout

`tsc -b` clean, `npm run lint` clean, `254 passed (853 assertions)`, PHPStan clean, Pint clean.

**The shipped bundle matches the source**: rebuilding this branch from scratch produces the identical content hash, `index-CBIplkO3.js`, and one entry chunk ships — the superseded one is deleted, as in #107.

---

*Correction to this description, found by taking the screenshots: the `search logout` row originally read `Routes 1` for the before state. It is `Routes 621` — the count read the unfiltered set, which is the whole point of the bug. The narrative below the table said so correctly; the table row did not.*


SanderMuller:

The sidebar with a query typed, in both states, is the whole story: `Routes 1 · Risks 2 · Recent 0` above a tree of one, next to `1 · 1 · 0`. Worth capturing the `Risks` tab with a filter on as well. The count disagreeing with the tree and the list showing unfiltered contents are two separate bugs, and the table asks a reader to hold both in their head at once.

webard:
Added — three pairs in the description: the `Routes` tab with the query, the `Risks` tab with the same query, and the chips narrowed to `DELETE`.

Captured with Playwright driving the sidebar rather than posed by hand, so the pair comes out of one run: same scan data, same viewport, only the commit differs (`80ad4c2` against `bf8151b`). Thank you for the pointer to eye-verification — I had the pieces for this already and went with a script in the same spirit; the repeatability is the part that mattered.

**Taking them found an error in my own description**, which is a good argument for your asking. The `search logout` row read `Routes 1` for the before state. It is **621** — the count read the unfiltered set, which is the entire bug. The narrative under the table had it right and the table row did not, and you quoted the wrong number back to me from it. Fixed, and noted at the bottom of the description rather than silently corrected.

You were also right that the `Risks` shot is the one the table could not carry: `/broadcasting/auth` sitting in a list filtered by `logout` says it in one look, and no row of numbers does.

One thing the `DELETE` pair might raise: the 125 is not 125 `DELETE` routes. Console commands, broadcast channels, schedules and Filament pages carry no HTTP method, so the chips do not apply to them and they stay in the tree. That is existing behaviour of the tree filter — this PR only makes the counts agree with it. Noted under the images so it does not read as a second bug.

Images live on a `pr-media` branch in my fork, so nothing lands in this diff.

SanderMuller (approved):
Reviewed at `bf8151bf`. This covers the whole diff, which is the six-commit stack: #99, #100, #101, #107, the build cleanup, and this fix. I reviewed the first four separately.

I could drive the real component this time rather than reason about it, so most of what follows is a run rather than a reading.

## Both bugs, reproduced on real scans

I rendered the actual `LeftSidebar` against three applications' manifests, on this head and on its parent, and counted what came out of the markup.

Unfiltered, which is where the second disagreement lives:

| application | before: badge vs cards under it | after |
|---|---|---|
| A | `Risks 27` above **85** cards | `Risks 85` above 85 |
| B | `Risks 37` above **79** cards | `Risks 79` above 79 |
| C | `Risks 34` above **76** cards | `Risks 76` above 76 |

The badge understated by between two and three times on every application I have, with no filter involved at all. Your description treats this as the secondary point; on this evidence it is the one a user was more likely to hit.

Then the headline bug, typing into the search box for real (jsdom, `react-dom/client`, an `input` event on the actual field):

```
BEFORE                                    AFTER
""       Routes 367 · Risks 37 · 79 cards  Routes 367 · Risks 79 · 79 cards
"DELETE" Routes 367 · Risks 37 · 79 cards  Routes  26 · Risks  7 ·  7 cards
"users"  Routes 367 · Risks 37 · 79 cards  Routes   4 · Risks  0 ·  0 cards
                                           + "No flagged routes match the filter."
```

Every badge frozen and the risk list untouched no matter what was typed, before. After, all three track the query and each badge equals the length of the list it labels, on both applications I tried. The new empty state appears exactly where a filter emptied the list.

One limit: `changeStatus` is absent from all three of my manifests (no previous scan), so `Recent` was 0 in every run and I did not exercise that path.

## The layout change that came along

`largestOf()` replacing five spreads is not just a tidy-up on my side, since #107's overlap measurements run through those functions. Re-run on this head: circle overlaps 0 of 537, 0 of 367, 0 of 423 tabs, grid 0 of 423, breadth-first 0 of 423, all unchanged. And the `RangeError` is gone: 65,536 / 124,000 / 200,000 nodes all lay out where 124,000 threw before.

## The bundle is the thing that ships, so I checked it reproduces

`npm ci && npm run build` on this head leaves `git status` empty. The committed chunk is byte-for-byte what the source produces, and the blade points at a chunk that exists. Deleting the three superseded entry chunks in `80ad4c20` is the right answer to the note on #107, and taking `index-Bze1eHA4.js` with them closes the loose end I flagged about #106's list.

CI is green for real here, which is worth saying because it has not been on the recent ones: 13 PHP jobs plus `frontend`, per-job results, not a "no checks reported" line. The output gate does not apply, and I did not run it: the whole diff is `frontend/`, the blade and built assets, with nothing under `src/` or `tests/`.

## One thing to know rather than fix

There is no frontend test framework in the repository, and the `frontend` job runs `npm ci`, `npm run lint`, `npm run build`. Nothing pins these counts. What I ran to check them took jsdom and an esbuild bundle, neither of which is in the project, and I would not ask you to add a test stack inside a bug fix. It is worth naming because the change makes it easy: the filter is now one memo over one set, and `filteredTabs`, `riskTabs` and `recentTabs` are pure functions of `(tabs, query, methods)` that a test could call directly if a runner ever lands.

Approving.

PR #109, opened 2026-08-31
Fix: two fingers on a trackpad resize the canvas instead of moving it

webard:
> **Stacked on #108**, which is stacked on #107, #99, #100 and #101. The last commit is this PR; the six below it are that stack. No other merged branch touches `frontend/src`.

## Two-finger scroll zooms

macOS sends a two-finger scroll as a wheel event like any other, and d3-zoom zooms on every wheel. The `deltaX` a trackpad reports was being read as scale, so **scrolling a graph sideways was not possible at all** — the gesture that means "look to the left" made the graph smaller.

A trackpad scroll now pans. A pinch and a mouse wheel still reach d3 and still zoom, so zooming stays anchored on the pointer exactly as d3 does it.

Deltas arrive in screen pixels while `translateBy` moves the graph's own coordinates, so they are divided by the current scale. Without that, panning drifts against the pointer at every zoom level except 100%.

## The part that needs your judgement

**Nothing in the event says which device sent it.** A notch of a mouse wheel and two fingers on a trackpad arrive the same shape. A pinch is the one case the browser marks, by setting `ctrlKey`. Everything else is inference, and the choice of what to infer is a product decision as much as a technical one:

- **Wheel pans, ctrl-wheel zooms** — what Figma, Miro and Excalidraw do. No inference needed, works on every device, and a mouse-wheel user loses zoom-on-scroll.
- **Infer the device** — a mouse keeps zooming and a trackpad pans, at the cost of a heuristic that will be wrong on some hardware.

This PR implements the second, because that is what the person who reported it wanted. **I would take a redirect to the first without argument** — it is less code, it cannot be wrong about a device, and the convention is well established. Say the word and I will swap it.

## The inference is measured, not assumed

389 real events captured from a MacBook trackpad in Chrome — two-finger scrolls, diagonal scrolls, pinches:

| gesture | events | what they carry |
|---|---:|---|
| scroll | 354 | `wheelDeltaY === -3 * deltaY` in 350, off by one unit in 4; `deltaX` frequently non-zero; `deltaY` whole; largest `\|deltaY\|` was **49** |
| pinch | 35 | `ctrlKey` set; `wheelDeltaY` pinned at exactly ±120 while `deltaY` stayed fractional and under 9 |

That `-3` ratio is the signal. It identifies a trackpad **positively**, rather than inferring one from the absence of a mouse — a notch reports `deltaY` around 100 against a `wheelDeltaY` of 120, a ratio of -1.2, nowhere near it.

The pinch row is also why `ctrlKey` has to be tested first rather than folded in: a pinch's `wheelDeltaY` is *always* a multiple of 120, so any rule keying on that alone reads every pinch as a mouse. I had written exactly that rule before measuring.

Fallbacks for what the capture could not cover: a non-zero `deltaX` (a wheel has one axis), a `deltaMode` other than pixels (mice on Firefox and Windows, never a trackpad), and a coarse whole `deltaY` where `wheelDeltaY` is absent (Firefox).

## What is not verified, and cannot be by me

**No mouse was available to test with.** Every path ending in "this is a mouse" is reasoned from Chrome's documented shapes, not measured. The known gap is written into the function: a high-resolution wheel reporting `deltaY` 40 against a `wheelDeltaY` of -120 hits the ratio exactly and would be read as a trackpad.

If you have a mouse to hand, logging `{deltaX, deltaY, deltaMode, wheelDeltaY}` from it is the one measurement that would close this, and `isTrackpadPan` is the only function to change.

## Verified

Measured event shapes replayed through the built handler, overriding `wheelDeltaY` because Chrome derives a different value for synthetic events than the hardware sends — which is itself worth knowing, since it means a naive synthetic test of this function passes for the wrong reason:

```
five trackpad shapes, including the rounding outlier  ->  pan,  5/5
two pinch shapes                                      ->  zoom, 2/2
three mouse shapes (unmeasured, from Chrome's docs)   ->  zoom, 3/3
```

And the panning itself, before the device split was added:

```
two-finger (60, 40)      ->  screen (-60, -40), scale unchanged
same at k = 0.18, 1.2, 5 ->  screen (-50, -25) every time
pinch                    ->  point under the cursor identical before and after
mouse drag (80, 60)      ->  screen (+80, +60), scale unchanged
```

The scale-independence is what asserts the division is right. The drag is there because the zoom filter changed and mouse panning goes through it.

`tsc -b` clean, `npm run lint` clean, `254 passed (853 assertions)`, PHPStan clean, Pint clean. Rebuilding the branch from scratch reproduces the shipped chunk's content hash; one entry chunk ships and the superseded one is deleted.


SanderMuller:
Could you add a recording to the description, before and after?

A still cannot show this one. A few seconds of a two-finger scroll on each side would do it: the graph shrinking when the gesture meant "look left", then panning. Your 389-event capture is a stronger argument than a video for whether the heuristic is right, but the video is what shows a reader what the bug felt like.

The device-inference question you raise is the decision worth reviewing, and I will come back to it properly.

webard:
Before:

https://github.com/user-attachments/assets/4357033b-4858-463e-96cf-808590e25378

After:

https://github.com/user-attachments/assets/fe6350f8-18dc-4ab4-9914-bab2033e8cc9

I’m using two fingers on the touchpad, which correctly pans the canvas instead of zooming it.

SanderMuller (approved):
Reviewed at `17f6213a`, covering the whole seven-commit stack. #99, #100, #101, #107 and #108 I reviewed separately; this is about the wheel handling on top.

## The classifier, run against every shape you measured

`isTrackpadPan` is module-private, so I added `export` to it, built, tested, then reverted and confirmed the tree was clean. Same function body, no other change.

```
  trackpad scroll, small           (deltaY 2,    wheelDeltaY -6)      -> pan
  trackpad scroll, largest seen    (deltaY 49,   wheelDeltaY -147)    -> pan
  trackpad scroll, upward          (deltaY -12,  wheelDeltaY 36)      -> pan
  trackpad, one-unit outlier       (deltaY 7,    wheelDeltaY -22)     -> pan
  trackpad diagonal                (deltaX 8, deltaY 5)               -> pan
  trackpad pure horizontal         (deltaX 11, deltaY 0)              -> pan
  pinch in                         (ctrlKey, wheelDeltaY -120)        -> zoom
  pinch out                        (ctrlKey, wheelDeltaY 120)         -> zoom
  mouse notch down                 (deltaY 100,  wheelDeltaY -120)    -> zoom
  mouse notch up                   (deltaY -100, wheelDeltaY 120)     -> zoom
  mouse, line mode (deltaMode 1)                                      -> zoom
  mouse, page mode (deltaMode 2)                                      -> zoom
  mouse, no wheelDeltaY (deltaY 120)                                  -> zoom
  shift-wheel sideways             (deltaX 40, deltaY 0)              -> pan
```

Fourteen for fourteen against what you describe, including the rounding outlier and the pinch that any `% 120` rule would misread. Testing `ctrlKey` first is doing the work you say it is.

**Your disclosed gap is real, and I could not find a second one.** A high-resolution wheel reporting `deltaY 40` against `wheelDeltaY -120` is classified as a trackpad, exactly as the docblock warns. The same family: `deltaY 0.5` against `wheelDeltaY -120` also pans. Both are the ratio test being permissive when a wheel reports finely, and both are covered by what you already wrote. I went looking for a shape you had not thought of and did not find one.

One curiosity, harmless: a wheel event with `deltaX`, `deltaY` and `wheelDeltaY` all zero classifies as a pan, so it calls `preventDefault()` and translates by nothing. Nothing moves either way. Not worth code.

## The pan arithmetic

Checked against d3 rather than by algebra on my side. `t.translate(-dx / t.k, -dy / t.k)` for a (60, 40) delta:

```
k=0.18   screen shift (-60.000, -40.000)   scale unchanged
k=1      screen shift (-60.000, -40.000)   scale unchanged
k=1.2    screen shift (-60.000, -40.000)   scale unchanged
k=5      screen shift (-60.000, -40.000)   scale unchanged
```

Exactly your claim, at every zoom level, with `k` untouched. The pinch path still goes through d3's own handler, so zoom stays anchored on the pointer, and `{ passive: false }` is required for the `preventDefault()` you make.

## The rest

`npm ci && npm run build` on this head leaves `git status` empty: the committed chunk is byte-for-byte what the source produces, and the blade points at a chunk that exists. CI is green with per-job results, 13 PHP jobs plus `frontend`. The output gate does not apply and I did not run it, since nothing under `src/` or `tests/` is touched by the whole stack. As on #108, there is no frontend test runner, so nothing pins this classifier against a future edit, and `isTrackpadPan` is the most testable function in the stack: a pure predicate over a plain object, with 389 real events already captured to test it with.

## The decision you asked for is not mine

You ask whether to keep the device inference or switch to wheel-pans / ctrl-zooms. That is the maintainer's call, so I have flagged it rather than answered it. What I can say from running both halves: the inference is correct on every measured shape, its failure mode is a mouse that pans instead of zooming rather than anything destructive, and the docblock tells the next person exactly where to look and what to log. If the answer is "switch", the diff gets smaller and this review's evidence stops being needed. Nothing here blocks either way.

Approving on correctness.

PR #114, opened 2026-08-31
Support Route::match([...], ...) in RouteAnalyzer

### LINKED ISSUE #113: brain:scan silently skips Route::match([...], ...) routes
## Bug: `Route::match([...], ...)` routes are silently skipped by `brain:scan`

**Package:** `laramint/laravel-brain` v2.6

### Description

`RouteAnalyzer::extractRoutes()`'s AST visitor only recognizes these HTTP verbs as route-defining calls:

```php
// src/Analysis/RouteAnalyzer.php
private const HTTP_METHODS = ['get', 'post', 'put', 'patch', 'delete', 'options', 'any'];
```

`match` is not in this list, and there is no separate branch for it in `enterNode()` (unlike `resource`, `apiResource`, and `livewire`, which each get their own handling). As a result, any route declared with:

```php
Route::match(['get', 'post'], '/example', [ExampleController::class, 'index']);
```

is completely invisible to `brain:scan` — it isn't counted in the route total, doesn't appear in the lifecycle graph, and isn't reflected in the Filament/command chain analysis either, since those all key off the same route extraction.

### Reproduction

1. In a fresh Laravel app, add to `routes/web.php`:
   ```php
   Route::get('/a', fn () => 'a');

   Route::match(['get', 'post'], '/b', fn () => 'b');
   ```
2. Run `php artisan brain:scan`.
3. Observe the route count reports `1 route`, not `2` — `/b` is dropped entirely.

### Expected

`Route::match([...], $uri, $action)` should be treated like `Route::get`/`Route::post`/etc., expanding to one `RouteDefinition` per method in the array (mirroring how Laravel's router itself registers it), including support for the same prefix/middleware/group context and post-chain methods (`->middleware()`, `->name()`, etc.) that `HTTP_METHODS` routes already get.

### Impact

For apps whose API routes are declared via `Route::match` (e.g. because a single endpoint intentionally accepts both `GET` and `POST`), `brain:scan` can undercount routes by a large margin and the generated graph misses those endpoints' controllers, middleware, and downstream call chains entirely.

Happy to send a PR adding `match` support to `HTTP_METHODS` handling (parsing the array-of-methods first arg and looping) if that's welcome.

xplodman:
## Summary

`Route::match([...], $uri, $action)` is not in `RouteAnalyzer::HTTP_METHODS` and has no dedicated branch in `enterNode()`, so `brain:scan` silently drops every route declared with it — no route count, no lifecycle/graph entry, no Filament/command chain tracing, nothing.

Fixes #113.

## Change

- Added `handleMatchRoute()`, which reads the methods array (`Route::match`'s first argument), and delegates to `handleHttpRoute()` once per verb.
- Gave `handleHttpRoute()` an `$argOffset` parameter, since `match()`'s URI/action arguments are shifted by one position compared to `get()`/`post()`/etc.
- Wired `match` handling into both the `StaticCall` and `MethodCall` branches of `enterNode()`, alongside the existing `resource`/`apiResource`/`livewire` special-casing.
- Also handled `match` as the base call inside `tryHandlePostChainedRoute()`, so `Route::match([...], '/uri', $action)->middleware(...)` style post-route chaining works too.

## Test plan

- Added a new test (`expands Route::match into one route per HTTP verb`) covering: multiple HTTP verbs, a wrapping `prefix()`/`middleware()` group, and post-chain `->middleware(...)`.
- `vendor/bin/pest tests/Unit/RouteAnalyzerTest.php` — all 26 tests pass.
- `vendor/bin/pest --order-by=random` — full suite passes except 4 pre-existing, unrelated failures in `DatabaseGraphStoreTest` caused by a missing `pdo_sqlite` driver in my local PHP 8.4 build (reproducible on `v2.x` before this change too).
- `vendor/bin/pint --test` on the touched files — passes.

### DIFF
--- a/src/Analysis/RouteAnalyzer.php
+++ b/src/Analysis/RouteAnalyzer.php
@@ -1224,12 +1224,22 @@ class RouteAnalyzer
                 if ($methodsValue === null) {
                     return;
                 }
-                $methods = $this->extractMiddlewareList($methodsValue);
+
+                // extractMiddlewareList() resolves a ClassConstFetch to its class name, which is
+                // right for `SomeMiddleware::class` but wrong for a verb written as a class
+                // constant (e.g. Request::METHOD_GET) — that would come back as the class's FQCN,
+                // not a verb. Filtering to the verbs this analyzer actually understands drops such
+                // an entry instead of registering a bogus route, matching how an unrecognised verb
+                // was already handled before Route::match() support existed: silently ignored.
+                $methods = array_values(array_intersect(
+                    array_unique(array_map('strtolower', $this->extractMiddlewareList($methodsValue))),
+                    [...self::HTTP_METHODS, 'head'],
+                ));
                 if ($methods === []) {
                     return;
                 }
 
-                foreach (array_unique(array_map('strtolower', $methods)) as $method) {
+                foreach ($methods as $method) {
                     $this->handleHttpRoute($node, $method, $extraMiddlewares, 1);
                 }
             }
--- a/src/Analysis/RouteAnalyzer.php
+++ b/src/Analysis/RouteAnalyzer.php
@@ -1075,6 +1075,8 @@ class RouteAnalyzer
 
                     if (in_array($methodName, self::HTTP_METHODS, true)) {
                         $this->handleHttpRoute($node, $methodName);
+                    } elseif ($methodName === 'match') {
+                        $this->handleMatchRoute($node);
                     } elseif ($methodName === 'group') {
                         $this->enterGroupFromStaticCall($node);
                     } elseif (in_array($methodName, ['resource', 'apiResource'], true)) {
@@ -1095,6 +1097,8 @@ class RouteAnalyzer
                         $this->enterGroupFromMethodChain($node);
                     } elseif (in_array($methodName, self::HTTP_METHODS, true)) {
                         $this->handleHttpRoute($node, $methodName);
+                    } elseif ($methodName === 'match') {
+                        $this->handleMatchRoute($node);
                     } elseif (in_array($methodName, ['resource', 'apiResource'], true)) {
                         $this->handleResource($node, $methodName);
                     } elseif ($methodName === 'livewire') {
@@ -1139,10 +1143,13 @@ class RouteAnalyzer
             /**
              * @param  string[]  $extraMiddlewares  Middleware collected from post-route chaining
              *                                      (e.g. Route::get(...)->middleware('ability:...'))
+             * @param  int  $argOffset  Index of the URI argument. 0 for Route::get('/uri', ...);
+             *                          1 for Route::match(['get','post'], '/uri', ...), whose first
+             *                          argument is the methods array rather than the URI.
              */
-            private function handleHttpRoute(Node\Expr\StaticCall|Node\Expr\MethodCall $node, string $method, array $extraMiddlewares = []): void
+            private function handleHttpRoute(Node\Expr\StaticCall|Node\Expr\MethodCall $node, string $method, array $extraMiddlewares = [], int $argOffset = 0): void
             {
-                $uri = $this->extractString($node->args[0] ?? null);
+                $uri = $this->extractString($node->args[$argOffset] ?? null);
                 if ($uri === null) {
                     return;
                 }
@@ -1159,10 +1166,10 @@ class RouteAnalyzer
                 $stackController = end($this->controllerStack) ?: '';
                 $controllerContext = $chainController !== '' ? $chainController : $stackController;
 
-                [$controller, $actionMethod, $closureNode, $mayPrependNamespace] = $this->extractAction($node->args[1] ?? null, $controllerContext);
+                [$controller, $actionMethod, $closureNode, $mayPrependNamespace] = $this->extractAction($node->args[$argOffset + 1] ?? null, $controllerContext);
 
                 // An array action can carry its own middleware alongside `uses`.
-                $actionArg = $node->args[1] ?? null;
+                $actionArg = $node->args[$argOffset + 1] ?? null;
                 $actionValue = $actionArg instanceof Node\Arg ? $actionArg->value : $actionArg;
                 if ($actionValue !== null) {
                     $actionMiddleware = $this->arrayValueFor($actionValue, 'middleware');
@@ -1200,6 +1207,31 @@ class RouteAnalyzer
                     closureNode: $closureNode,
                     closureUseMap: $closureNode !== null ? $this->useMap : null,
                 );
+            }
+
+            /**
+             * Handles Route::match(['get', 'post'], '/uri', $action) — registers one
+             * RouteDefinition per HTTP verb named in the methods array, mirroring how
+             * Laravel's router itself expands a match() call.
+             *
+             * @param  string[]  $extraMiddlewares  Middleware collected from post-route chaining
+             *                                      (e.g. Route::match([...], ...)->middleware('auth'))
+             */
+            private function handleMatchRoute(Node\Expr\StaticCall|Node\Expr\MethodCall $node, array $extraMiddlewares = []): void
+            {
+                $methodsArg = $node->args[0] ?? null;
+                $methodsValue = $methodsArg instanceof Node\Arg ? $methodsArg->value : $methodsArg;
+                if ($methodsValue === null) {
+                    return;
+                }
+                $methods = $this->extractMiddlewareList($methodsValue);
+                if ($methods === []) {
+                    return;
+                }
+
+                foreach (array_unique(array_map('strtolower', $methods)) as $method) {
+                    $this->handleHttpRoute($node, $method, $extraMiddlewares, 1);
+                }
             }
 
             /**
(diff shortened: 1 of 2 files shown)


SanderMuller (approved):
Reviewed at `bea91f13`. The defect is real and the fix works on a production application. Two notes below, neither blocking.

## What it recovers, measured

I ran the analyzer over three applications, before and after. One of them declares three `Route::match(['get', 'post'], …)` routes, and the output gate shows what they were hiding:

```
application A   identical
application B   74 ids added, 0 removed
application C   identical
```

Nothing is lost anywhere. What appears on B is 6 route nodes, 10 actions, a controller, a middleware, a view, six new tabs and the edges between them: a whole login flow the graph could not see, because the three declarations that reach it were skipped. A null arm on each application came out identical, so those are the change's numbers and not the machine's.

## Every shape I could think of

A fixture with eleven `match` declarations, run on both arms. Before: 2 routes, the two that were not `match`. After: 17.

| written as | after |
|---|---|
| `match(['get', 'post'], …)` | GET + POST |
| `match(['GET', 'DELETE'], …)` | GET + DELETE, lowercased on the way in |
| `match('get', …)`, the string form Laravel also accepts | one GET route |
| `match(['get', 'GET'], …)` | one route, not two |
| `match([], …)` | nothing, no crash |
| `match(['get', 'head'], …)` | GET + HEAD, a verb outside `HTTP_METHODS` |
| a closure action | GET + POST |
| `->middleware(…)->name(…)` after the call | both verbs carry the middleware |
| inside `prefix()->middleware()->group()` | both verbs, prefix applied, group and chained middleware both present |
| `Route::any` and `Route::get` beside it | unchanged |

No duplicates anywhere, which is the thing I most expected to find: the post-chain path returns `DONT_TRAVERSE_CHILDREN`, so the inner call is never visited twice. Reusing `extractMiddlewareList()` for the methods array is what makes the string form work for free.

`406 passed (1134 assertions)`, PHPStan clean, Pint clean. Your four `DatabaseGraphStoreTest` failures are environmental, as you said: with `pdo_sqlite` loaded they pass here.

CI has not run on this branch. It reports `no checks reported`, which a fork PR needs a maintainer to release, so that green is mine, local, on `bea91f13`.

## The new test does not pin the case in the issue

This is the note worth acting on. Delete the `match` branch from the `StaticCall` arm of `enterNode()`, the one that handles a plain `Route::match([...], '/uri', $action)` with nothing chained after it, and **all 26 tests still pass**. My fixture goes from 17 routes to 6 with that same edit: every unchained declaration disappears.

The test's fixture only exercises the chained form, and the chained form is handled by `tryHandlePostChainedRoute()` instead, so the branch that covers the shape the issue is about is the one nothing is watching. One more line in the fixture fixes it:

```php
Route::match(['get', 'put'], '/plain', [\App\Http\Controllers\AuthController::class, 'plain']);
```

Symmetrically: deleting one of the two `match` branches inside `tryHandlePostChainedRoute()` also leaves the suite green, because the other one covers the test's shape.

## A verb written as a class constant becomes the class

```php
Route::match([Request::METHOD_GET], '/constant', [AuthController::class, 'constant']);
```

comes out as a route whose method is `ILLUMINATE\HTTP\REQUEST`. `extractMiddlewareList()` resolves a `ClassConstFetch` to its class reference, which is exactly right for `SomeMiddleware::class` and wrong for a verb constant, and `handleMatchRoute()` uppercases whatever it gets.

It is narrow. None of the three applications writes a verb that way, and before this PR such a route was dropped entirely rather than mis-registered, so nothing regresses. Worth a guard anyway, because `SecurityAnalyzer` keys its mutating-method checks on the exact strings `POST`, `PUT`, `PATCH` and `DELETE`, so a `Request::METHOD_POST` route would be recorded as neither mutating nor GET. Filtering the expanded verbs to the ones `HTTP_METHODS` knows, plus `head`, would drop it back to the old behaviour of ignoring what it cannot read.

## Verdict

Approving. The gate difference is pure addition on a real application, the expansion matches what Laravel's router does with `match()`, and nothing else in the output moved. Both notes are follow-ups: the fixture line is the one I would do before merge, since without it the branch that fixes #113 is untested.

One mechanical thing for whoever merges: this targets `v2.x`. The two branches happen to be identical right now, so it makes no difference today. Worth a look before merging in case they have diverged again by then.

xplodman:
Thanks for the thorough review @SanderMuller — pushed a follow-up commit (`2404cd4`) addressing both notes:

1. **Test gap**: added a top-level, unchained `Route::match([...], '/plain', $action)` to the fixture (no group, no post-chain), asserting on both expanded verbs. Verified locally that reverting just the `StaticCall` branch of `match` handling in `enterNode()` now correctly fails this test (it previously stayed green).

2. **Class-constant verb**: `handleMatchRoute()` now filters resolved verbs down to `HTTP_METHODS + ['head']` before registering routes, so `Route::match([Request::METHOD_GET], ...)` is silently dropped (matching the pre-fix behavior for any unrecognised verb) instead of being registered with method `Illuminate\Http\Request`. Added a test covering both the pure-constant case and a mixed literal+constant case.

Full suite: 406 passed / 4 pre-existing unrelated failures (missing `pdo_sqlite` in my local PHP 8.4), Pint clean, PHPStan clean on the touched file.

SanderMuller (approved):
Re-reviewed at `2404cd4`, against `bea91f1` — the head my last review described. Both notes are addressed, and I re-ran the probes that produced them rather than reading the diff.

**The fixture line does what it needed to.** Deleting the `match` branch from the `StaticCall` arm of `enterNode()` — the one handling a plain unchained `Route::match([...], '/uri', $action)` — now fails two tests. On the previous head all 26 passed. That was the branch fixing the issue, and it is now the branch a test is watching.

**The class-constant guard behaves exactly as the old code did for a verb it cannot read.** Same fixture, both heads:

| | `bea91f1` | `2404cd4` |
|---|---|---|
| `Route::match([Request::METHOD_GET], '/constant', …)` | `ILLUMINATE\HTTP\REQUEST` | dropped |
| `Route::match(['get', Request::METHOD_POST], '/mixed', …)` | `GET, ILLUMINATE\HTTP\REQUEST` | `GET` |
| `Route::match(['get', 'head'], '/withhead', …)` | `GET, HEAD` | `GET, HEAD` |

The mixed row is better than what I asked for — the readable verb survives and only the unreadable one is dropped, rather than the whole route going. And allowing `head` through the filter keeps the existing behaviour intact, which the third row confirms.

Suite green locally: **407 passed (1144 assertions)**, PHPStan no errors, Pint clean.

## Two things about the other three `match` branches

I mutated all four separately this time. Two fail a test, two do not, and the two that do not mean opposite things — so I chased both rather than reporting them together.

**`Route::middleware('auth')->match([...], '/uri', …)` is handled and nothing watches it.** Deleting the `MethodCall` arm's `match` branch leaves the suite green, and that shape then disappears entirely:

```
as shipped                     /prechain  GET,PUT
MethodCall branch deleted      /prechain  — MISSING —
```

Same shape of gap as the one you just closed, one branch further along, and the same one-line fix: a `Route::middleware(...)->match(...)` declaration in the fixture.

**The first of the two post-chain `match` branches appears to do nothing.** Deleting it leaves the suite green *and* leaves the output unchanged across every shape I could construct — plain, post-chained, pre-chained, both, `->name()->middleware()` doubled, inside a group, and pre-chained inside a group. Seven shapes, identical results with and without it. The second post-chain branch answers all of them. I could not find the input only the first serves, which is not the same as proving there isn't one, but it is worth a look before this merges.

## The thing whoever merges needs to know

**CI still has not run on this branch.** `gh pr checks` reports `no checks reported`, which a fork PR needs a maintainer to release. The green above is mine, local, on `2404cd4`.

**#128 implements the same feature, is already approved, and conflicts with this.** `git merge-tree` on the two heads reports a conflict in `src/Analysis/RouteAnalyzer.php`; they solve `Route::match` by different mechanisms — a dedicated `handleMatchRoute()` with four call sites here, a `resolveRouteVerbs()` folded into `handleHttpRoute` there. Run against the same fixture, they agree everywhere except one row:

| | this branch | #128 |
|---|---|---|
| `[Request::METHOD_GET]` | dropped | dropped |
| `['get', Request::METHOD_POST]` | `GET` | `GET` |
| `Route::middleware(…)->match(…)` | `GET, PUT` | `GET, PUT` |
| plain unchained | `GET, PUT` | `GET, PUT` |
| `['get', 'head']` | `GET, HEAD` | **`GET`** |

#128 drops HEAD deliberately, because Laravel registers it alongside GET on its own and the live-router path already skips it — so a HEAD tab duplicates the GET one. I verified that reasoning when reviewing it.

That makes #128 a superset of this branch's behaviour, so this is a maintainer's call rather than a defect: whichever merges second needs its conflict resolved by hand, and merging both would leave two implementations of one feature in the same file. If #128 goes in, the useful thing left here is the test coverage, not the code.

Also still true from last time: this targets `v2.x`, which is level with `main` today. Worth re-checking at merge time.

## Verdict

Approving, and re-pointing the approval at this head — the previous one described `bea91f1`. The two notes I raised are fixed and measured. The overlap with #128 is not a reason to hold this branch; it is a reason for somebody to decide which of the two ships.

PR #116, opened 2026-09-01
Feat: read each model's real schema, and flag foreign keys nothing indexes

webard:
> Independent of #115 — branched from `main`, no dependency either way. Both add `ModelTableResolver` as a byte-identical file, which merges without conflict.

A model node described the class and never the table. Its columns, its indexes and its foreign keys — the things you need before writing a query against it — were not in the graph at all.

Model nodes now carry the live schema behind a **Schema** tab, and a foreign key with nothing to read it by is raised as a finding in the same channel, with the same badge and colour, as a route without authentication.

| the schema | the finding |
|---|---|
|  |  |

<details>
<summary>Columns, indexes and foreign keys in one tab</summary>

</details>

## From the catalogue, not from migrations

Migrations say what someone intended when they wrote them. Any project of age has a schema that no longer matches their sum — an index added during an incident, a column a later file altered, a staging database that drifted from production. The catalogue is the only account of what is actually there.

It is also the portable one. `getColumns()`, `getIndexes()` and `getForeignKeys()` return identically-shaped rows on all five drivers Laravel supports, so **this change contains no SQL and no branch on driver**. Measured: 150 tables, three calls each, **0.97s**.

## The finding is about reads, not writes

A foreign key constrains what may be written. On PostgreSQL and SQL Server it creates no index at all — MySQL's InnoDB is the exception, and creates one implicitly, so the finding is naturally quiet there and loud where it matters. Every `belongsTo` traversal, `whereHas` and cascading delete through an unindexed key scans the child table: instant on seed data, pathological on real data.

On the application this was built against: **10 unindexed keys across 9 of 151 models**, including `location_id` and `lot_id` on the inventory ledger, which is the largest table in the system. Nine of a hundred and fifty-one is the ratio that matters — a check that flagged most models would be one people learn to scroll past.

Coverage is a **leading-prefix** test, not a set test. An index on `(order_id, created_at)` covers a key on `order_id`; one on `(created_at, order_id)` does not, because the engine cannot seek into it by its second column. Both directions are pinned, and swapping the comparison for a set intersection kills two tests.

**Severity stays at `medium`.** A missing index is a real defect and a slow one, but it is not an unauthenticated write, and levelling the two would teach people to skim the badge that exists to stop them skimming.

## Three things this turned up

- **Model nodes were getting their payload only where a `ModelDefinition` was passed** — and three of the four call sites that create one pass `null`: a related model, an observed model, one reached from a channel. `InventoryMovement` had its schema on the ERD tab and on none of the 34 route tabs it also appears on. Keyed by FQCN now: **0 route-tab nodes became 34**.
- **The findings panel was gated on `isRoute`**, so a model had nowhere to show them.
- **The exposure chip would have fallen back to `Public`** for any value it did not recognise, labelling a table as publicly exposed — a claim, and a frightening one. To be precise about what this is: it is *not* a defect users are seeing today. On `main` a model node never reaches that code, because `securityData` is gated on `isRoute`; the fallback becomes reachable only once findings are shown for non-routes, which is the change above. So the guard prevents something this PR would otherwise introduce rather than fixing something already broken. Exposure is a route's concept; an unrecognised one draws no chip at all, and the exposure card is skipped rather than announcing a "Public Route" for a database table.

## Opting out, and what an unreachable database costs

`laravel-brain.schema.enabled` turns the step off for a purely static scan; `schema.connection` points it at another connection when the models do not live on the default one. Failure is quiet throughout — no connection, no permission, an unreadable table all end as a missing schema and a finished scan.

Quiet is not the same as quick, though, and the first version of this got that wrong. A refused connection fails instantly, but a host that drops packets — a database reached over a VPN that is not up — blocks for PDO's default of **30 seconds**, on every scan and on every poll in watch mode.

`schema.timeout` (default **2 seconds**) now bounds it. Measured against a host that drops packets:

| driver | no timeout | `timeout = 2` |
|---|---|---|
| pgsql | 30.02s | **2.00s** |
| mysql | 30.01s | **2.00s** |

The bound is applied through `PDO::ATTR_TIMEOUT` on a *copy* of the connection, so the application's own connection keeps the settings it was configured with, and the connection is opened eagerly in `forConnection()` — `DB::connection()` resolves lazily, so leaving the connect to the first query would have carried the wait into the middle of the scan where no bound applies.

One measurement worth recording, because it contradicts the usual advice: `PGCONNECT_TIMEOUT` was built first, on the belief that PDO_PGSQL ignores `ATTR_TIMEOUT`. Measured on this pairing it is the other way round — `ATTR_TIMEOUT` alone bounds pgsql at **2.01s**, and `PGCONNECT_TIMEOUT` alone leaves it at **30.01s** — so that mechanism was removed as dead weight rather than kept as belt-and-braces.

## Verified

Eight unit tests on the coverage rule, including the composite cases in both directions, both mutation-checked. Checked in a browser against a real 623-route application: the tab renders on route tabs as well as the ERD tab, and the finding appears in the Risks tab, on the node's badge, and on the card in the canvas.

`413 passed (1134 assertions)`, PHPStan clean, Pint clean, `tsc -b` and `eslint` clean. One entry chunk and one stylesheet ship; the superseded ones are deleted.


SanderMuller (changes requested):
Reviewed at `e4d5dca9`. This is good work and I verified most of it end to end against a real database. One thing needs changing before merge, and it is about time rather than correctness.

## The blocking one: an unreachable database costs 30 seconds, every scan

`schema.enabled` defaults to true, so every existing user's scan now opens a connection. Failure is quiet, as you say, but quiet is not the same as quick. I measured the two ways a connection fails:

```
connection refused (a stopped container on localhost)   0.02s
host drops packets (a remote database, VPN off)        30.01s
```

Thirty seconds is PDO's default connect timeout. A scan that takes one to two seconds on the applications I measure with becomes thirty-one, with `→ Reading database schema...` on screen and no indication that anything is wrong. The step sits in `runAnalysis()` with no early return before it, and `analyze()` is the only entry point, so **watch mode pays it again on every poll**.

The refused case is instant, which covers "docker is down". The thirty-second case is "my database is on a host I reach over the VPN, and the VPN is off", which is an ordinary Tuesday.

What would change my mind: a bounded connect timeout on the connection the inspector opens. I tested that it works before suggesting it, using the same unreachable host with `PDO::ATTR_TIMEOUT => 2` in the connection options:

```
host drops packets, 2s timeout                          2.01s
```

A `schema.timeout` beside `schema.connection`, defaulting to a couple of seconds, would keep the failure both quiet and cheap. Anything that bounds the wait is fine by me; I am not asking for a particular shape.

## What I verified, and it holds

**The catalogue read, end to end on a real driver.** I built a sqlite database with four foreign keys and ran `SchemaInspector` against it:

| the key | the index situation | flagged |
|---|---|---|
| `order_id` | `(order_id, created_at)` | no |
| `other_id` | only `(created_at, other_id)` | **yes** |
| `third_id` | none | **yes** |
| `mixed_id` | index declared `MIXED_ID` | no |

The leading-prefix rule does exactly what your docblock says, in both directions, and a column named in a different case than the index that covers it is still matched because the catalogue normalises before you see it. A table that does not exist is skipped without an error.

**The mutation you named kills the tests you say it kills.** Replacing the sequence comparison with `array_intersect($columns, $index['columns']) === $columns` fails `still flags a composite index that does not lead with the key` and `flags a composite key whose columns appear in the wrong order`, and nothing else. Eight tests, and those two are the ones carrying the rule.

**Output gate: identical on three applications, which is the result I wanted and not a pass to wave through.** None of them has a database configured in my harness, so the schema step reads nothing, and the graph is byte-for-byte what it was. That is the no-database path proven not to disturb anything, on 21k, 31k and 49k node graphs. A null arm on each came out identical, so those are the change's numbers.

**The viewer.** The committed bundle reproduces: `npm ci && npm run build` leaves `git status` empty, one entry chunk and one stylesheet are added and their predecessors deleted, and the blade names files that exist. Rendering the real `Sidebar` against a model node carrying a finding and a schema: `Schema` and `Risks` tabs both appear, and no exposure chip is drawn for an unrecognised value. The same node on `main` gets neither tab. Two of your three collateral fixes, confirmed by running them.

`413 passed (1134 assertions)`, PHPStan clean, Pint clean. CI is genuinely green here: 15 jobs with per-job results, including the new Benchmark one.

## One smaller thing

The exposure chip is the one claim I could not reproduce as a *pre-existing* defect. On `main` a model node never reaches that code, because `securityData` is gated on `isRoute`, so the `?? exposureColors['public']` fallback could not fire for a table. It becomes reachable only once findings are shown for non-routes, which is this PR. The guard is right and I would keep it exactly as written. It fixes something this change would otherwise have introduced, rather than something users are seeing today, and the description reads as though it is the latter.

## Verdict

Requesting changes for the connect timeout alone. Everything else I looked at is sound: the rule is right, its tests bite, the no-database path is provably inert, the bundle matches its source, and the finding is pitched at a severity that will not train people to ignore the badge. Put a bound on the wait and I will approve on the re-review.

webard:
Thank you — this is the most useful kind of review: you reproduced the thing, measured it, and told me what would change your mind. Both points are addressed in `908d59f`.

## The blocking one: bounded

`schema.timeout`, default 2 seconds, applied through `PDO::ATTR_TIMEOUT`. Your numbers reproduce exactly on my side, and the bound does what you measured:

| driver | no timeout | `timeout = 2` |
|---|---|---|
| pgsql | 30.02s | **2.00s** |
| mysql | 30.01s | **2.00s** |

Two details beyond the bare option, both of which turned out to matter:

**The connection is now opened eagerly in `forConnection()`.** `DB::connection()` resolves lazily, so the connect was happening at the first query inside `inspect()`. Setting the option without forcing the connect would have left the wait exactly where it was, just with an option attached to it. Opening it eagerly also means an unreachable database turns the feature off cleanly rather than carrying a dead connection into the middle of the scan.

**The timeout goes on a copy of the connection**, registered as `laravel-brain-schema`, not on the connection itself. A scan runs inside the booted application, and quietly shortening the connect timeout on the connection the rest of it uses is a side effect nobody asked for.

## One measurement that contradicts the usual advice

I first built this with `PGCONNECT_TIMEOUT` alongside the attribute, on the well-repeated claim that PDO_PGSQL ignores `ATTR_TIMEOUT` and libpq must be told through the environment. Isolating the two against a host that drops packets says the opposite:

```
pgsql, ATTR_TIMEOUT=2 only         2.01s
pgsql, PGCONNECT_TIMEOUT=2 only   30.01s
```

So the environment variable bounds nothing here and was removed rather than kept as belt-and-braces — along with the two tests and the save/restore dance around it. The docblock records the measurement, because the next person to read that code will have heard the same advice I did.

Three tests cover the mechanism without needing a network: the bound lands on the copy and not on the original, no timeout configured leaves no copy at all, and an unopenable connection yields no inspector. Removing `ATTR_TIMEOUT` from the derived config turns the first one red. Gate: `pint` clean, `phpstan` clean, **416 passing**.

## The smaller thing: you are right, and the description was wrong

The exposure chip is reachable only *because* of this PR — `securityData` is gated on `isRoute` on `main`, so a model node never reaches the fallback. I had it filed as a pre-existing defect and it is not one. The description now says plainly that the guard prevents something this change would otherwise introduce, rather than fixing something users are seeing today. The guard itself is unchanged, as you suggested.

Thanks also for re-running the mutation and the leading-prefix table against a real sqlite database — the `MIXED_ID` case in particular is the one I would have most expected to have got wrong.

SanderMuller (approved):
Re-reviewed at `908d59f`, against `e4d5dca` — the head my last review described.

I asked for a bounded connect timeout and said any shape that bounds the wait would do. `908d59f` bounds it, and the measurement is the same one that produced the finding.

**The probe, re-run against a host that drops packets:**

| | wall |
|---|---:|
| connection refused, nothing listening | 0.01 s |
| packets dropped, no bound (the driver's own default) | **30.01 s** |
| packets dropped, `schema.timeout` = 2 | **2.01 s** |
| packets dropped, `schema.timeout` = 5 | 5.01 s |

The number tracks the setting, which is what separates a bound from a host that happened to answer quickly today. The unbounded row is the control, run in the same session on the same address.

**The copy really is a copy.** After the call, the application's own connection carries `options: null` and only `database.connections.laravel-brain-schema` carries `PDO::ATTR_TIMEOUT`. That is the part I would have worried about — a scan quietly shortening the timeout on the connection the rest of the application uses — and it does not happen.

**The eager `getPdo()` is load-bearing, not defensive.** Removing it fails `it reads no schema rather than failing when the connection cannot be opened`; applying the bound to the real connection instead of a copy fails `it applies the connect timeout to a copy, leaving the real connection alone`. Two mutations, two named tests, each the one you would want.

**A suspicion that dissolved, worth recording since it was the one thing I thought this commit might cost.** Forcing the connect at construction means a project with no models at all now opens a connection it has nothing to ask. I expected that to be a small regression, and measured it on both heads with zero tables and the unreachable host:

```
e4d5dca   forConnection 0.00s, then inspect([]) 30.01s
908d59f   forConnection 2.01s, then inspect([])  0.00s
```

The old code paid the full thirty seconds for zero tables too — it just paid it later, inside `inspect()`, which is worse rather than better. So the eager connect costs nothing that was not already being spent, and this head improves the empty case along with every other. If you want it free rather than cheap, skipping the whole block when `$tables === []` would do it, but that is a nicety now, not a fix.

**Output gate: identical on all three applications**, with a null arm run first. None of them has a database configured in my harness, so this proves the no-database path is still inert rather than proving anything about the schema read — same caveat as last time, stated for the same reason.

Suite green on this head: **416 passed (1140 assertions)**, PHPStan no errors, Pint clean, 15/15 CI.

The note in the config block is the right length and says the right thing, and the docblock recording that `PGCONNECT_TIMEOUT` bounds nothing on this driver pairing while `ATTR_TIMEOUT` does — with the two measurements next to it — is the kind of thing that saves the next person an afternoon. Leave it in.

Approving.

mrmarchone:
Hello @webard , could you fix the conflicts ?

webard:
Conflicts resolved in `fd8502e` — rebased onto `main` at `af41e62`.

The collision is with **#115**, which landed while this was open: table statistics and this branch's schema read are two independent readings of the same tables, they stamp the same ERD node, and they were inserted at the same points in seven files. Both are kept everywhere; nothing was dropped in favour of the other.

## The three that needed a decision rather than a concatenation

**The model table resolver now runs once for both.** Each side called `(new ModelTableResolver)->resolve($models)` inside its own `if`, so a naive merge would have parsed every model twice to answer the same question. It is hoisted above both steps and guarded by `tableStatsEnabled || schemaEnabled`.

**`buildErdTab()` takes both.** Signature is now `(…, array $tableStats = [], array $schemas = [])`, and the node is written in **one** `updateNodeData` call carrying whichever of the two is present — `updateNodeData` replaces rather than merges, so two sequential writes would have had the second drop what the first added. A model may have either, both or neither.

**One behaviour I deliberately did not change.** #116 resolves a model's table through `tableByFqcn` before falling back to the definition, because three of the four paths that create a model node pass no definition at all; #115 reads it from the definition only. Giving statistics the same fallback would put them on many more nodes — plausibly an improvement, but a change to a merged feature, and not one to make silently inside a conflict resolution. Each side keeps the lookup it shipped with, and there is a comment at that spot saying so.

## The bundle

Resolved by rebuilding rather than by picking a hash: `npm run build` regenerates the blade and the entry chunk, and the two chunks the conflict left behind are removed. That rebuild is also what caught a real mistake — my first pass at `types/graph.ts` merged the two conflicting sides *inside* an interface body and produced invalid TypeScript. `tsc -b` refused it; a text-level resolution would have shipped it.

Gate on the rebased branch: `pint` clean, `phpstan` no errors, **443 tests passing**, and CI is green on all 15 jobs.

PR #118, opened 2026-09-01
Feat: event choreography, and the listeners each graph sets off

webard:
Adds an **Events** tab showing every event the application defines, what listens to it, and what those listeners fire in turn — and carries that knowledge onto every other graph that shows an event.

Events were only ever known to the graph when a traced call chain happened to dispatch one. Measured on a real modular application: **45 of 211**. The 108 events nothing listens to — the most interesting ones on the list — were invisible by construction, because an event's whole point is that its consumers are not on the dispatcher's path.

Discovery is by directory, the convention Laravel's own generator writes to, and configurable via `laravel-brain.events.paths`. Being filed under `Events/` is treated as the declaration it is, rather than inferred from a name or an interface — events implement no common one. `events.enabled` turns the whole thing off.

## Deferral is read from all three places that decide it

Reading one of them produces confident nonsense:

- `ShouldDispatchAfterCommit` on the **event**
- `ShouldQueue` on the **listener**
- `after_commit` on the **queue connection**

The last is a `queue.php` setting that no amount of reading the class will find. On the application measured, **all eight connections set it** — so every queued listener there is already deferred and only synchronous ones can act early. A check that looked only at the event class reported 6 problems where 2 are real.

<table>
<tr>
<td></td>
<td></td>
</tr>
<tr>
<td>Immediate dispatch, one synchronous listener — it can act before the commit.</td>
<td>The case a class-level check cannot see: nothing on the event or the listener says this, only <code>after_commit</code>.</td>
</tr>
</table>

`observableBeforeCommit` is reported as a **precondition, deliberately not as a finding**. It says what happens *if* an event is dispatched inside a transaction; it does not claim that it is. Answering that needs the transaction boundaries themselves (#117), and reporting it as a problem without them would raise one on every event that merely *could* be dispatched that way — 52 of 211 here. The rule becomes worth having once both are in.

## The facts follow the event onto every graph

A route graph reaches an event the moment the request dispatches one, and the node stopped at name-and-file. One route showed eight events, one of which sets off **seven** listeners and two of which set off nothing at all — and the graph said neither.

So the facts are computed once and stamped onto whatever graph carries the node, rather than assembled inside the tab that happens to be about events. Stamping merges into the node's existing data: `updateNodeData` replaces the whole array, so writing one key would drop the file path and the flow steps — silently, since a node with no file still renders. There is a test pinning exactly that.

## Listeners now appear on the graphs that dispatch the event

A listener edge is the one hop whose caller is a class nobody calls: the event. Given the generic treatment its caller id became `app_events_orderitemcreated::__construct`, while every graph that shows the event shows `event::FQCN` — **two nodes for one class**, with the listener hanging off the one no route can reach. The listener nodes existed in the full graph all along and appeared in no tab: **102 route tabs carried an event node and not one carried a listener.** This is the same fix the model observers already use.

77 route tabs now show what their events set off — 410 listener nodes in all, at most **+13** on any one tab. They are ordinary nodes, so they arrive with flow steps, source and the existing N+1 analysis; on the route above, deleting an order item turns out to run a synchronous listener that queries in a loop.

`listener` also joins the **Show on graph** legend with its own colour and its own toggle, so the nodes can be counted and hidden like any other type. Without that entry they rendered but appeared nowhere.

## Wide layers wrap

A layer's width was bounded by nothing but how many nodes sit at that depth. A graph of mostly unconnected entry points laid them in a single row kilometres wide, readable only at a zoom where no label is — the Events tab opened at **3%**. Layers over twelve nodes now wrap into a block; at or under it, layout is unchanged (verified on the route tabs).

## Verification

`pint`, `phpstan` and the suite are green — **432 passing**, 18 new. New rules were mutation-checked rather than assumed: reverting the `after_commit` term, the promoted-property read, the abstract-event skip, the known-event filter on the second hop, the canonical listener wiring, and the merge-not-replace in stamping each turn a test red.

One test I wrote did *not* discriminate — a guard for "no listeners" that the empty loop already produced — so the guard was removed rather than kept as decoration.

There is no frontend test runner in this repo, so the layer wrapping was verified in the browser across three tab shapes (Events, Model ERD, and a normal route tab) rather than by test.


SanderMuller (changes requested):
Reviewed at `7ca64c4e`. The Events tab and the deferral reasoning are good, and the layer wrapping does exactly what you say. One thing blocks: on one of my three applications, three route tabs **lose** the listeners they used to show.

## Three route tabs lose a listener and gain nothing

The listener edge now hangs off `event::FQCN`. The node the *dispatching* chain puts on a route tab is still the old one, because `nodeIdForHop` is untouched for the hop that constructs the event. So both nodes exist, and a tab that has only the pseudo-node loses the listener that used to hang off it.

Before, in one tab:

```
app_events_someevent::__construct  -[action-to-listener]->  app_listeners_somelistener::handle
```

After, in the full graph:

```
event::App\Events\SomeEvent        -[event-to-listener]->  app_listeners_somelistener::handle
```

and that tab still contains the `::__construct` form, not the `event::` one. The listener is no longer reachable from the tab, so it is gone from it.

Measured across three applications, tabs that lose a node with nothing replacing it:

| | tabs losing nodes | nodes lost |
|---|---:|---:|
| A | 0 | 0 |
| B | **3** | **5** |
| C | 0 | 0 |

The five are all listener handlers, on a login route and two registration routes (names changed here, shape exact). In that application's full graph, two event classes still exist in both forms at once, which is the same "two nodes for one class" the commit sets out to remove:

```
app_events_someevent::__construct              <->  event::App\Events\SomeEvent
illuminate_auth_events_registered::__construct <->  event::Illuminate\Auth\Events\Registered
```

What would change my mind: resolve the dispatch hop to the canonical id too, so a route tab carries `event::FQCN` and the listener travels with it. That is what the observer wiring does, and it should also deliver the 77-tab result on applications shaped like this one, where today it delivers a loss instead. Any approach that leaves no tab worse off is fine by me.

## The rest of the output, classified

One new tab per application (`events--choreography`), and nothing else changes shape:

| | ids added | ids removed | existing tabs whose node count moved |
|---|---:|---:|---|
| A | 53 | 13 | none |
| B | 66 | 26 | 3, all losses, above |
| C | 35 | 10 | none |

Everything removed is a `*_events_*::__construct` pseudo-node or an edge attached to one, on all three. Everything added is `event::` nodes (31 / 20 / 19), `listener::` nodes (5 / 15 / 5), their edges, and the tab. Null arms were identical, so those are the change's numbers.

## What I could not exercise, and why that is not evidence against it

The headline of the third commit, listeners appearing on the route tabs that dispatch the event, does not reproduce here, and the reason is the applications rather than the code. On C, 27 existing tabs carry an event node, but only 5 of its 25 discovered events have a listener at all, and none of those 5 appears on any route tab. There is nothing for the feature to add. Worth saying so rather than reporting a silent gate as a verdict either way.

## Verified

**The wrapping boundary is exactly where you say.** Driving `layoutBreadthFirst` with one layer of n nodes:

```
11 nodes -> 1 row,  span 2,450px
12 nodes -> 1 row,  span 2,695px     at or under the threshold, unchanged
13 nodes -> 3 rows, span 1,225px
25 nodes -> 4 rows, span 1,470px
```

No overlapping cards at any size. The kilometre-wide row is gone and the threshold does not fire early.

The committed bundle reproduces: `npm ci && npm run build` leaves `git status` empty, one entry chunk replaces its predecessor, and the blade names a file that exists. `432 passed (1159 assertions)`, PHPStan clean, Pint clean, CI green across 15 jobs with per-job results.

The `after_commit` reasoning is right and worth the trouble: reading only the event class is a check that reports problems an application has already solved in `queue.php`, and the no-container path returning "deferred" is the correct direction to fail in, so a scan without a booted app invents nothing.

## Verdict

Requesting changes for the lost listeners. Everything else here holds up, and the parts I expected to be shaky (the wrapping threshold, the pseudo-node collapse, the deferral rule) are the parts that came out cleanest.

webard:
Fixed in `42f56a5`. You were right, and the measurement on my side is worse than yours before the fix and better after it.

## The dispatch hop now resolves to the canonical id

Both id maps — `hopCalleeNodeId()` and `ensureNode()` — send an `event` hop to `eventId()`, so the node a dispatching chain puts on a route tab is the same node the listeners hang off. The node that hop builds is also renamed from `Event@__construct` to just the class name: the id no longer separates one method from another, so a label naming one would claim a distinction the node does not make.

Measured on a 60-module application, which reproduces exactly the shape you found:

| | before | after |
|---|---:|---:|
| canonical `event::` nodes | 601 | 622 |
| legacy `::__construct` event nodes | 21 | **0** |
| classes existing in **both** forms | **17** | **0** |
| tabs showing a listener | 78 | **91** |
| listener nodes | 562 | **586** |

So on that application the branch was losing listeners the same way, on a larger scale than application B — and after the change thirteen tabs **gain** them. Node and edge totals move the way merging predicts: 4033 → 4018 and 10295 → 10287.

## Why nothing caught it, which is the part I want to fix properly

The tests either side of the seam both build their own input. `EventsTabTest` assembles the tab from hand-made `EventDefinition`s, and the listener edges are constructed directly — so an event with two node ids satisfied both while no route tab held both ends.

The fixture this now runs against has the shape none of them had: a route, a controller that dispatches, an event, and a listener registered for it. Two assertions on it — the class resolves to exactly one node id, and the listener is present in that route's own subgraph. Reverting either map entry turns both red.

The events fixture had no routes and no controller at all, which is precisely why this was invisible; that seemed worth saying out loud rather than just fixing.

Gate: `pint` clean, `phpstan` no errors, **434 passing**.

Thank you for running it across three applications — a defect that shows on one of three is exactly the kind I would have shipped on a green suite and a single-application check.

SanderMuller (approved):
Re-reviewed at `42f56a5`, against `7ca64c4` — the head my last review described.

I asked for the dispatch hop to resolve to the canonical id so a route tab carries `event::FQCN` and the listener travels with it, and said any approach that left no tab worse off would do. That is what `42f56a5` does, and the measurement says it does exactly that.

**The probe that produced the finding, re-run — per tab, keyed by tab id, never by position.** Null arm first: the previous head against itself, identical on all three applications.

| | tabs compared | gaining listeners | **losing listeners** | listeners across all tabs |
|---|---:|---|---:|---|
| A | 538 | 0 | **0** | 5 → 5 |
| B | 371 | 3 (0→2, 0→1, 0→2) | **0** | 15 → 20 |
| C | — | byte-identical | — | — |

Three tabs on B gain the listeners they were missing, and no tab anywhere loses one. That is the result the previous head inverted.

**Nothing in the listener wiring moved to pay for it.** Listener-touching logical edges are 7 → 7 on A and 15 → 15 on B, with none lost and none gained, and no listener node is left without an incoming edge on either side. The merge changes which node the dispatch points at, not what points at the listeners.

**Both new branches are pinned, mutated one at a time.** Reverting the `calleeNodeId` case fails `it keeps the listener reachable from the tab of the route that dispatches it`; reverting the other case fails that one and `it puts the dispatched event on one node, not two`. Two lines, two branches, checked separately rather than together.

Suite green on this head: **434 passed (1163 assertions)**, PHPStan no errors, Pint clean, 15/15 CI.

## One thing the re-key does that is worth knowing about

Keying an event node on the FQCN alone is right for an event, and the comment saying so is right too. It is lossy for anything the tracer has *mistyped* as an event, because those nodes did carry a real per-method distinction.

Across three applications this happens once. On A, five nodes collapse into one:

```
before   ...\Console\Commands\<AConsoleCommand>::<methodA> | ::<methodB> | ::<methodC> | ::<methodD> | ::<methodE>
after    event::App\Console\Commands\<AConsoleCommand>
```

Five distinct method hops on one console command, merged into a single node whose surviving `data.method` is whichever arrived first. The class is not an event at all — it is a console command whose name happens to contain the word "Events", which is why the tracer typed the hops that way. It is typed that way on the previous head too, so the misclassification is not yours. The re-key just turns a harmless mislabel into four lost nodes on one command tab.

Every other merge on both applications is the intended one: three `__construct` duplicates on A and five on B, each folding into the canonical node. Net event nodes go 39 → 35 on A and stay at 37 on B.

Not blocking, and I would not hold the fix for it. Two small things if you want to close it off: the label correctly stopped naming a method, but `data.method` still carries one of the five, which is the same claim the label was fixed to stop making. And a guard that only re-keys when the class actually looks like an event — under an `Events\` namespace, or implementing the framework's contract — would keep the merge where it belongs.

Approving.

PR #119, opened 2026-09-01
Fix: stop calling every property read in a loop an N+1

webard:
The N+1 marker was mostly noise. `FlowExtractor::isQuery()` answered yes for **any** `PropertyFetch`:

```php
if ($expr instanceof Node\Expr\PropertyFetch) {
    $obj = $this->shortExpr($expr->var);   // computed, never read
    // If it's $this->relation or $item->relation, it's often an N+1
    return true;
}
```

That `$obj` is the leftover of a condition nobody wrote. And because a method call falls through to its receiver, `$this->service->handle()` inside any loop registered as a query — a shape that appears in almost every loop. Measured on a real application: **28 of 33 markers came from that one branch.**

## What it asks now

A relation read is an N+1 when it is read **off the value the loop is iterating**. The fetch chain is followed back to its root variable, and that root has to be bound by an enclosing `foreach`:

| in a loop | before | now |
|---|---|---|
| `$order->customer` (loop binds `$order`) | N+1 | N+1 |
| `$order->customer->address` | N+1 | N+1 |
| `$this->releaseClaim->execute($row)` | N+1 | — |
| `$this->config->timeout` | N+1 | — |
| `Warehouse::query()->find($id)` | N+1 | N+1 |

The loop's variable is popped when the loop ends. The extractor is reused across every method it charts, so a name left bound goes on flagging reads of that name in the next loop and in the next method — there is a test for exactly that, and it was written because a mutation of the pop survived the first round of tests.

## Applications that autoload relations

`Model::automaticallyEagerLoadRelationships()` loads the relation for the whole collection on first touch, so even the genuine shape is not an N+1 there and the branch is skipped.

It arrived in **Laravel 12.8** and this package supports 9 through 13, so the call itself is the version gate: on anything earlier there is no such method and the attempt throws, which is the same answer as "off" — an application that cannot turn it on has not turned it on. A `method_exists` guard would read better but cannot be written honestly, because static analysis runs against one installed version where the method always exists.

Asking the framework also beats scanning providers for the call, which would miss it being set from a package, a config flag, or a conditional.

An explicit query in a loop is untouched by any of this: batching relation access does not batch a query somebody wrote out.

## Measured

On the same application, end to end:

| | markers |
|---|---|
| before | **33** |
| after narrowing to the loop variable | **7** |
| after also honouring relation autoloading | **6** |

The narrowing does the work; autoloading removes one more here. Survivors were spot-checked and are real — for instance a command calling `$destination->subscriptionTenants()->get()` and `forceDelete()` once per row, neither of which autoloading batches.

## Verification

`pint`, `phpstan` and the suite are green — 413 passing, 7 new tests covering both directions of each rule. All three rules were mutation-checked: restoring "every property fetch is a query", ignoring autoloading, and dropping the loop-variable pop each turn a test red. The third only did so after a test was added for it — the first mutation of the pop survived, which is what surfaced the cross-method leak.

### DIFF
--- a/src/Analysis/FlowExtractor.php
+++ b/src/Analysis/FlowExtractor.php
@@ -27,7 +27,22 @@ class FlowExtractor
 
     private array $useMap;
 
-    public function __construct()
+    /**
+     * Variable names bound by the foreach loops currently being descended into, innermost last.
+     *
+     * A relation read is only an N+1 when it is read off the thing the loop is iterating. Without
+     * this the heuristic counted every property fetch, so `$this->service->handle()` inside any
+     * loop registered as a query — the single largest source of false markers.
+     *
+     * @var list<string>
+     */
+    private array $loopVars = [];
+
+    /**
+     * @param  bool  $relationsAutoloaded  Whether the application batches relation access, which
+     *                                     makes a relation read inside a loop not an N+1 at all.
+     */
+    public function __construct(private readonly bool $relationsAutoloaded = false)
     {
         $this->printer = new PrettyPrinter;
         $this->useMap = [];
@@ -209,13 +224,16 @@ class FlowExtractor
             $expr = $this->shortExpr($stmt->expr);
             $value = $this->shortExpr($stmt->valueVar);
             $label = "foreach ({$expr} as {$value})";
+            $this->loopVars[] = $this->rootVariableName($stmt->valueVar) ?? '';
             $body = $this->stmtsToSteps($stmt->stmts, true);
+            $n1 = $this->hasQueryInside($stmt->stmts);
+            array_pop($this->loopVars);
 
             return [
                 'type' => 'loop',
                 'label' => $label,
                 'body' => $body,
-                'n1' => $this->hasQueryInside($stmt->stmts),
+                'n1' => $n1,
             ];
         }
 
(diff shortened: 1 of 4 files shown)


SanderMuller (approved):
Reviewed at `6b2e0a8d`. The diagnosis is right, the narrowing works, and I could not find a genuine finding it drops. Approving, with two notes.

## What it removes, and what it keeps

I ran the extractor on both arms over the same shapes:

| in a loop | before | after |
|---|---|---|
| `$order->customer` | flagged | flagged |
| `$order->customer->address` | flagged | flagged |
| `Warehouse::query()->find($id)` | flagged | flagged |
| `$this->service->handle($order)` | **flagged** | not |
| `$map[$order->id] = 1` | not | not |
| a relation read after the loop closes | not | not |

Then over three applications, end to end:

| | flow steps carrying `n1` | tabs with `n1Count > 0` |
|---|---|---|
| A | 791 → **474** | 48 → 35 |
| B | 116 → **98** | 15 → 13 |
| C | 109 → **54** | 15 → 10 |

Node and edge ids are identical on all three, so nothing moved but the markers. Null arms were identical.

## No true positive lost, checked properly

My first pass at this compared marker labels and looked like it had found eleven lost findings on one application. Every one dissolved on inspection, and the reason is worth writing down: the old rule flagged the *enclosing loop* and its *sibling statements* whenever a single `$this->` read appeared anywhere in the body, so a nested loop over a relation carried a marker it had not earned. Reproduced exactly:

```php
foreach ($questions as $question) {
    foreach ($question->attributes as $attribute) {
        $this->mapper->record($attribute);      // before: three markers, one per line
    }                                            // after:  none
}
```

Take the `$this->` line out and the base flags nothing either. So all three markers came from that one branch, which is your 28-of-33 measured on a different application.

## Two things worth a follow-up

**A destructured loop variable loses detection.** This one is a regression rather than a gap:

```php
foreach ($pairs as [$order, $meta]) {
    $c = $order->customer;      // before: flagged. after: not.
}
```

`rootVariableName()` returns null for a list destructuring, so `?? ''` pushes an empty name and nothing inside that loop can match. Pushing each variable the pattern binds rather than one name would close it. It is narrow, and I would not hold the PR for it.

**The shape the rule is named for is not detected on either arm.** A loop over a relation of the outer loop's variable is the textbook N+1:

```php
foreach ($questions as $question) {
    foreach ($question->attributes as $attribute) { … }   // neither arm flags this
}
```

The loop's own iterated expression never reaches `isQuery()`; only its body does. That predates this PR and is not something it makes worse, but `rootVariableName()` is exactly the helper that would answer it, so it is a small follow-up rather than a redesign.

## Everything else

All three rules bite, and each kills only what it should: restoring "every property fetch is a query" fails three tests, removing the autoloading branch fails one, and dropping `array_pop($this->loopVars)` fails the cross-method leak test alone. Good instinct writing that last one: the leak is the kind that shows up as a wrong marker three methods later.

The reflection probe reads `isAutomaticallyEagerLoadingRelationships()`, a getter, so a scan asks the question without answering it: nothing about the analysed application changes because it was scanned. The catch-all returning "off" is the right direction, since it leaves the heuristic at its historical strength rather than silently disarming it.

`413 passed (1130 assertions)`, PHPStan clean, Pint clean, CI green across 15 jobs with per-job results.

## Verdict

Approving. A change that deletes four fifths of a marker's output needs to show it deleted the right four fifths, and this one does. I went looking for the lost finding specifically, and the only thing I found was the destructuring case, which is a line of code rather than a rethink.

PR #120, opened 2026-09-01
Feat: show what a queued job promises when it goes wrong

webard:
A job node carried its name, its file and its flow — so "this runs on a queue" was the whole story. Not whether a failure is retried, not whether a second dispatch is dropped on the floor, not whether the job refuses to overlap with itself. Those are the differences that matter at three in the morning.

<table>
<tr>
<td></td>
<td></td>
</tr>
</table>

The section appears only when the job declares something. A job that declares none of it gets no section at all — six nulls say less than nothing.

## Two sources, for two kinds of fact

Interfaces and the `Batchable` trait are asked of the **loaded class**, because `ShouldBeUnique` is as often inherited from a base job as written on the job itself.

Everything else is read from the **source**. A declared `$tries = 5` is a fact the file states, and instantiating a job to ask it is not something an analyzer should do — a constructor can take models, open connections, or throw.

## A fact expressed as a method is still a fact

Reading properties alone would have found a third of it. Measured on an application of 113 jobs:

| declared as | count |
|---|---|
| `ShouldBeUnique` | 44 |
| `uniqueId()` | 32 |
| `middleware()` | 14 |
| `retryUntil()` | 8 |
| `$timeout` | 8 |
| `$tries` | 7 |
| `backoff()` | 4 |

So a method is reported too — as its **value** where the body is a single `return 60;`, and otherwise as **decided at runtime**. A computed backoff cannot be reduced to a number by reading the source, and printing a guess would be worse than printing nothing.

Middleware names come from the `new` at the **root** of each entry. Middleware is habitually configured by chaining, and the outermost node of

```php
new WithoutOverlapping($key)->releaseAfter(60)->expireAfter(180)
```

is a method call, not the class anyone means. A bare `SomeMiddleware::class` entry is read too.

## Scope

This enriches the job nodes a graph already has; it does not discover jobs. On the application measured that is 27 of 113 — the rest are not reachable from any traced call chain, which is a separate question from this one.

`Bus::chain()` and `Bus::batch()` are deliberately left out. They are call-site facts rather than class facts, and showing a chain properly means grouping its members and drawing the order between them — the region machinery from #117. Once that is in, a chain can reuse it rather than grow a second copy. The class-level half of it, the `Batchable` trait, is included here.

## Verification

`pint`, `phpstan` and the suite are green — 413 passing, 8 new. Three rules were mutation-checked: dropping the literal-return read, taking the outer node of a middleware chain instead of its root, and letting a job that declares nothing through to the panel each turn a test red.

`Batchable` reported zero across the measured graph, so rather than trust the zero it was checked directly against a job known to use the trait, which reports `batchable: true`. None of the reachable job nodes happens to be one of the four that use it.


SanderMuller (approved):
Reviewed at `aa537823`. The facts it reads are the right ones and the reading is careful. Approving, with two small things.

## What it reads, run on shapes rather than read

```php
class MethodFactsJob {
    public $tries = 5;
    public $timeout = 120;
    public function backoff(): int { return 60; }
    public function retryUntil(): \DateTime { return now()->addMinutes(5); }
    public function uniqueId(): string { return 'x'; }
}
```

comes out as `tries=5  timeout=120  backoff=60  dynamic=[retryUntil, uniqueId]`. The literal return becomes a number, the two that cannot be reduced say so instead of guessing, and neither is silently dropped. That is the distinction the description rests on and it holds.

On real job classes across three applications: 26, 40 and 29 described, of which 6, 15 and 9 declare anything at all. So the panel appears on roughly a quarter to a third of jobs, which is the right shape for a section people are meant to read rather than scroll past.

## `::class` inside a middleware entry is read as a middleware

```php
public function middleware(): array
{
    return [
        (new WithoutOverlapping(Order::class))->releaseAfter(60)->expireAfter(180),
        new RateLimited('reports'),
        SomeBareMiddleware::class,
    ];
}
```

reports `[WithoutOverlapping, RateLimited, Order, SomeBareMiddleware]`. `Order` is the key handed to the overlap guard, not a middleware.

The `new` scan does what the docblock says. The `::class` scan beside it walks the whole method body rather than the top level of each returned entry, so any class constant used as an argument comes back as a name. The bare `SomeMiddleware::class` form it exists for is the top-level case, so restricting it to entries of the returned array keeps that and drops this.

It is narrow: of the 16 `middleware()` methods across three applications, none passes a `::class` as an argument, so this is a shape I constructed rather than one I found. Worth a follow-up rather than a block.

## The superseded entry chunk is left behind

The branch adds `index-BwWD5GvV.js` and does not delete the one it replaces, so both ship and only the new one is referenced. #117, #118 and #119 each deleted their predecessor; this is the one that does not. One `git rm` and the count stays at one.

## Everything else

**Gate:** node and edge ids are identical on all three applications and the null arms came out identical, so the enrichment adds data to job nodes and moves nothing else. That is what an enrichment should look like.

**The tests bite, each on its own axis.** Dropping the literal-return read fails `it takes the value from a method whose body is just a return`. Taking the outer node of a middleware chain instead of the `new` fails `it names the middleware class, not the chained configuration around it`. Making `isInteresting()` always true fails `it says nothing at all about a job that declares nothing`. One test each, nothing collateral.

**Loading classes is guarded properly.** `class_exists()` inside a `try/catch (Throwable)` covers the case that matters: a job whose parent or interface is missing raises an `Error` rather than a warning, and that is caught. Asking the loaded class for `ShouldBeUnique` and the trait is the right call given both are as often inherited as written.

`413 passed (1132 assertions)`, PHPStan clean, Pint clean, CI green across 15 jobs with per-job results. The committed bundle reproduces from source.

Also worth saying: catching your own fixture for using syntax only 8.4 parses, in a second commit, is the kind of thing the matrix exists to find and usually finds later.

## Verdict

Approving. The `::class` case and the leftover chunk are both a line each, and neither is a reason to hold a change that turns "this runs on a queue" into something you can act on at three in the morning.

PR #121, opened 2026-09-01
Fix: record the container registrations the analyzer was dropping

webard:
`ContainerBindingAnalyzer` was dropping most of the container registrations written in front of it, and one branch downstream could never fire because of it.

## Measured

Distinct abstracts recorded, same provider paths both sides:

| Codebase | Before | After |
|---|---|---|
| Illuminate source (`vendor/laravel/framework/src`) | 100 | **167** |
| A 60-module Laravel application | 60 | **86** |

40% and 30% missing. Two causes, both confirmed by reading the code before changing it:

- `count($args) < 2` returned early, so `$this->app->singleton(Foo::class)` — an ordinary self-binding — was never recorded.
- The abstract was resolved with `resolveClassLike()`, which requires a namespace separator. Laravel's own bindings are routinely keyed by bare alias: `'mailer'`, `'cache'`, `'db'`. 61 of the 67 newly-found Illuminate registrations are exactly that.

## The unreachable branch was real

`FacadeRegistry::resolveWith()` only inspects facades whose accessor has **no** namespace separator — with one, `FacadeAnalyzer` sets `concreteFqcn` itself and `resolveWith` skips the facade. The registry only ever held keys **with** a separator. The two sets were disjoint by construction, so that branch could not fire on any input the two analyzers produce together.

Probed on a project with `singleton('short-url', ShortUrlService::class)` and a facade returning `'short-url'`: the facade resolved to `NULL` before and to `App\Services\ShortUrlService` after. It now resolves any application facade fronting a container alias — which is what the accessor-as-string branch of `FacadeAnalyzer` was written to feed in the first place.

The pre-existing test for that path hands the registry a `ContainerBindingRecord` built by hand — one that nothing could produce from source. It is left in place, and an end-to-end test now runs both analyzers over one fixture. That test asserts its own premise first (the accessor really is a bare key, `concreteFqcn` really is null) so it cannot pass vacuously.

## Widening paid for with tightening

The receiver check (`$this->app`, `$app`, `app()`) is the only thing separating a registration from any other `bind()`/`singleton()` call, so it was not relaxed — it was leaned on:

- **Single-argument form takes class-shaped arguments only.** A bare `->bind('x')` is the shape an unrelated one-argument `bind()` most often has. This costs nothing measurable: across both codebases every single-argument registration was class-shaped and none was bare.
- **The concrete position stays FQCN-only.** `bind(Foo::class, 'reporting')` is legal alias chaining, but `GraphBuilder` builds a node from `concreteFqcn`, and an alias there would materialise a class that does not exist. The abstract is recorded; the concrete is left null.
- **Alias charset** `^[A-Za-z0-9_][A-Za-z0-9_.-]*$`, derived from the 61 keys Illuminate actually uses. An out-of-shape key is missed, never mis-recorded.
- **Named arguments are read by name.** `bind(concrete: X, abstract: Y)` was previously recorded **backwards** — a false entry rather than a missing one, which is the worse defect of the two, and one that already occurs in the wild.
- **Spread and first-class-callable give up on the whole call.** `bind(Foo::class, ...$rest)` has an unreadable concrete; recording just the abstract would file it as a self-binding the source contradicts.
- **Self-bindings set concrete = abstract**, matching `Container::bind()`'s own `if (is_null($concrete)) { $concrete = $abstract; }`. Null already means "closure, unknown".

A `DecoyServiceProvider` fixture holds every one of those shapes, and a test asserts that **no** record carries that provider's name.

## No graph output changed

The obvious risk of widening a registry is churn in the graph. There is none, and it was measured rather than assumed: a second worktree at `main` produced `GraphBuilder` edge sets for all 18 pre-existing fixtures, compared byte for byte against this branch — identical, `GraphBuilderTest`'s exact `toHaveCount(63)` included.

The reason is structural: `maybeWireContainerBinding()` only fires when the abstract is an interface or an abstract class. Bare aliases are never a `calleeFqcn`, and self-bindings of concrete classes are not abstractions, so neither can add an edge. No existing assertion needed to move.

## Deliberately left out

`Container::bind()` also routes a one-argument **closure** through `bindBasedOnClosureReturnTypes()` — `singleton(static fn (): Generator => …)` binds `Generator`. It is real and genuinely single-argument, but reading it means resolving union and intersection return types into a set of abstracts, and it appeared twice across both codebases against 31 of the `X::class` form. Recorded as a decision in the source docblock so it reads as considered rather than missed.

## Verification

`pint`, `phpstan` and the suite are green — **417 passing**, 12 new, no existing test changed. Ten mutations were run and all ten killed: restoring the single-argument rejection, self-binding concrete to null, routing the single-argument form through the wrong resolver, resolving the abstract with `resolveClassLike`, dropping the charset rule, widening the concrete to aliases, reading named arguments positionally, removing the spread bail-out, class-only keys in the `$bindings` array, and disabling the receiver check.

### DIFF
--- a/src/Analysis/ContainerBindingAnalyzer.php
+++ b/src/Analysis/ContainerBindingAnalyzer.php
@@ -20,6 +20,18 @@ use PhpParser\NodeVisitorAbstract;
 /**
  * Scans service providers for Laravel container registrations (bind/singleton/scoped
  * and $bindings).
+ *
+ * A registration is recorded when the receiver looks like the container
+ * ({@see isAppLikeInvokable()}) and the abstract position resolves. The two positions are
+ * resolved by different rules on purpose: the abstract is a container KEY, which may be a class
+ * name or a bare alias ({@see resolveContainerKey()}), while the concrete has to be a class the
+ * rest of the pipeline can locate a file for ({@see resolveClassLike()}).
+ *
+ * Measured over the Illuminate source and over one application of 60 modules, counting distinct
+ * records in the registry: the analyzer used to find 100 and 60, and now finds 167 and 86 — it
+ * was missing 40% and 30% of the registrations written in front of it. Of the 67 the framework
+ * scan gained, 61 are bare aliases and 6 are single-argument self-bindings; of the application's
+ * 26, 25 are single-argument self-bindings and 1 is a bare alias.
  */
 final class ContainerBindingAnalyzer
 {
@@ -186,7 +198,7 @@ final class ContainerBindingAnalyzer
             $keyExpr = $item->key;
             $val = $item->value;
             if ($keyExpr instanceof Expr) {
-                $abstract = $this->resolveClassLike($keyExpr, $namespace, $useMap);
+                $abstract = $this->resolveContainerKey($keyExpr, $namespace, $useMap);
                 $concrete = $val instanceof Expr\Closure ? null : $this->resolveClassLike($val, $namespace, $useMap);
                 if ($abstract !== null) {
                     $registry->add(new ContainerBindingRecord($abstract, $concrete, $providerFqcn, $kind));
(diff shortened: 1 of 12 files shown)


SanderMuller (approved):
Reviewed at `04e043d4`. I reproduced the widening on my own applications, checked that it only adds, and confirmed the two downstream claims by running both analyzers rather than reading them. No blocking finding.

## The widening, on three applications

Distinct records the analyzer produces, same provider paths both sides:

| | before | after | nothing removed | nothing changed |
|---|---:|---:|:---:|:---:|
| A | 7 | **14** | ✓ | ✓ |
| B | 20 | **32** | ✓ | ✓ |
| C | 20 | **21** | ✓ | ✓ |

Purely additive is the part that matters most, and it is worth stating separately: no key the old analyzer recorded came out differently, so nothing downstream can have been re-pointed.

The new records fall into exactly the two causes you name. On B, nine of the twelve are FQCN self-bindings, and that application's providers contain exactly **nine** occurrences of `$this->app->singleton(<Class>::class)`. The rest are alias-keyed: ten with a null concrete (a closure behind a bare key) and one alias to an FQCN.

Worth knowing what each half buys, since they are not the same. The alias-to-FQCN kind is what the facade path consumes. The alias-to-null kind cannot reach `GraphBuilder`, which needs a `concreteFqcn`, so its value is in the registry itself.

## The unreachable branch, run end to end

One fixture, `singleton('short-url', ShortUrlService::class)` and a facade returning `'short-url'`, both analyzers, both arms:

```
before   facade accessor=short-url   concrete=NULL
after    facade accessor=short-url   concrete=App\Services\ShortUrlService
```

Your account of why is right: the registry only ever held keys with a separator, and `resolveWith()` only inspects accessors without one, so the two sets were disjoint by construction.

## The backwards record, confirmed on the base

This is the one I most wanted to see for myself, because a false entry is worse than a missing one. Same fixture, `bind(concrete: ShortUrlService::class, abstract: 'named-arg-key')`:

```
before   binding App\Services\ShortUrlService -> NULL      the concrete filed as the abstract
after    binding named-arg-key -> App\Services\ShortUrlService
```

Exactly as described, and it was a real defect on `main` rather than a hypothetical.

## The tightening, run rather than asserted

Your decoy fixture holds a non-container receiver, a bare string key, a key with spaces, a spread and a first-class callable. Running the analyzer over that project produces **9 records, all from `BindingShapesServiceProvider` and none from `DecoyServiceProvider`**. The guard rails hold when exercised, not only when asserted.

Four of your ten mutations, sampled and each killing only its own test:

| mutation | fails |
|---|---|
| reject the single-argument form again | `it records a single-argument self-binding with the abstract as its own concrete` |
| read arguments positionally | `it binds named arguments by name…`, `it gives up on a registration whose arguments are spread` |
| drop the alias charset | `it rejects a string that cannot be a container key`, plus the receiver test |
| resolve the abstract with `resolveClassLike` only | four, including the end-to-end facade one |

The charset mutation failing the *receiver* test is a nice accident: accepting any string lets an unrelated `bind()` through, so the two guards are load-bearing together rather than separately.

## No graph output changed, on more than the fixtures

Byte-identical on all three applications, with node and edge id counts unchanged (49,292 / 21,054 / 31,225) and null arms identical. Your structural argument holds outside the 18 fixtures you checked it on: bare aliases are never a `calleeFqcn`, and a self-binding of a concrete class is not an abstraction, so neither reaches `maybeWireContainerBinding()`.

`417 passed (1153 assertions)`, PHPStan clean, Pint clean, CI green across 15 jobs with per-job results.

## Verdict

Approving. Widening a registry by a third while proving the graph did not move is the hard version of this change, and the tightenings are what make the widening safe rather than decoration around it. Leaving the closure-return-type case out with the reasoning in the docblock is the right call at two occurrences against thirty-one.

PR #122, opened 2026-09-01
Feat: list every scheduled task, with when it runs and what it sets off

webard:
The sidebar's `Schedules` bucket held **exactly one row**, labelled *Scheduled Tasks*, no matter how many tasks the application had. It could answer "does this app schedule anything?" and nothing else.

Each task is now its own row and its own tab: what it runs, when it runs, and the guards around it. Three fixed lines rather than a wrapped stack, because a command name and a row of guards are both open-ended and the panel is not.

## Four defects found while reading, each confirmed with a probe before it was fixed

**The legacy `Kernel::schedule()` form lost the cadence entirely.** There the registration sits *inside* the cadence call — `$schedule->command('x')->daily()` — so walking outward from the registration reaches only `$schedule`. All three entries in a probe kernel came back with `frequency=''`. No test covered that path; the repo had no kernel fixture at all.

**`timezone` was in `FREQUENCY_METHODS`**, and the chain was read outermost-first, so `->daily()->timezone('Europe/Warsaw')` was recorded as running on a frequency of *timezone*.

**The kernel visitor matched `command`/`job`/`call` anywhere in the file**, so `$this->app->call(...)` inside `schedule()` became a scheduled closure on no cadence.

**`bootstrap/app.php` → `withSchedule(…)` was not scanned at all.** A Laravel 11+ skeleton has no Console Kernel, so such an application showed no schedule whatsoever.

## Linking a task to what it runs

A command node is keyed by its whole signature, option definitions and all; a schedule names the command and nothing more. Matching one string against the other only ever succeeded for a command that declares no options — measured on a 60-module application, **4 of 59** tasks found their command and 55 drew as a lone node.

A command can be named three ways — by signature, by name, or by class, since `Schedule::command(Foo::class)` carries the FQCN — so all three now resolve through one lookup order. That takes the tasks showing nothing from **55 to 9**, and the nine left are honest: a scheduled vendor command has no node in a graph of *this* application, and a closure has no target at all.

## A task no longer draws as the thing it runs

The viewer reads a node's kind out of its data, and a schedule node carried a `type` key of its own holding the kind of its *target* — `command`, `job`, `call`. That overwrote the node's real kind, so a task scheduling a command rendered as a command: two identically-labelled cards stacked one above the other, joined by `runs`, one of which was the schedule. The target's kind is now `targetType`, which collides with nothing; the manifest payload driving the CMD/JOB/FN markers keeps its own `type` and is untouched.

## What a scheduled closure does

Every other kind of task points at a class the graph can chart. A `Schedule::call(fn)` points at nothing, so its tab was one node and silence — and it is often the task running every minute.

Its body is now read where the closure is already in hand, and the tracer descends from it with the schedule's own node id standing in for the calling class, exactly as a closure route is already traced. Measured: a closure calling a model yields a model node, one calling a job yields a job node. A closure calling something the tracer does not chart at all still yields none — that is the tracer's existing reach, not a decision made here.

## The panel stopped scrolling sideways

The tree was `width: max-content`, so one long route path made every row wider than the panel and the whole sidebar scrolled horizontally — reading the end of one row hid the start of all the others, and a task's guards sat past the fold, cut mid-word.

Each open-ended line now scrolls inside itself, against the sidebar's real width, which the component publishes as a custom property because the panel is draggable. Measured in the browser at the default width: the panel's scroll width equals its client width, 24 of 111 task lines scroll within themselves, and nothing — task line or route name — extends past the panel edge.

## Verification

`pint`, `phpstan` and the suite are green — **426 passing**, no existing test changed in a way that weakened it. Mutations run and killed across the work: `timezone` back in `FREQUENCY_METHODS`, the scheduler-scope check dropped, `withSchedule` discovery disabled, `nodeId()` no longer hashing cadence arguments, guards never recorded, the chain walk not stopping at the registration, the manifest dropping its `schedule` key, schedule-tab dedup removed, the command lookup reverted to full-signature matching, the class lookup removed, `commandName()` returning the whole signature, the closure reader returning nothing, and closure steps attached to non-closure tasks.

Two of those needed a fixture that could actually discriminate: the fixture closure was an empty stub, which would have passed any test written against it, and there was no kernel fixture at all.


SanderMuller (approved):
Reviewed at `55d18e9a`. The headline defect reproduces on all three applications I measure with, and the fix reads correctly on every one. Approving, with two notes.

## The cadence really was gone, and is not now

All three of my applications use the legacy `Kernel::schedule()` form, which is the path you say lost the cadence. Running the analyzer over one of them, before and after:

```
before   18 schedule entries, every one with frequency = ''
after    dailyAt x20 (arguments ['8:00'], ['3:00'], ['2:00'] …), weekly x2, everyFiveMinutes x2
```

`ScheduleEntry` had no `timezone` property at all before, so that half is new rather than moved. On a kernel written as `->daily()->timezone('Europe/Warsaw')`, the frequency comes back `daily` with the timezone beside it, which is the distinction the second defect was about.

## Nothing existing moved

The risk in re-keying schedule nodes is that a tab quietly loses something, so I compared every tab by id rather than by position:

| | tabs | new | gone | shared tabs losing nodes | gaining |
|---|---|---:|---:|---:|---:|
| A | 537 → 548 | 12 | 1 | **0** | 0 |
| B | 370 → 374 | 5 | 1 | **0** | 0 |
| C | 424 → 428 | 4 | 0 | **0** | 0 |

Every new tab is a `schedule-command-…` or `schedule-job-…`, one per task. The one that goes is `schedule--tasks`, the old single bucket, which is the intended replacement rather than a loss.

Worth one line for whoever merges: that is a tab id disappearing, so a stored scan keeps a `schedule--tasks` blob nothing addresses any more. #98's pruning covers it on the next scan; it is only a note because ids are the thing stored scans are keyed by.

C is the interesting one. It had no schedule tab at all before and now has four, all jobs, and none of them draws an edge, because none of those four job classes exists as a node anywhere in that graph. That is the reach limit you describe rather than the linking failing, and I checked it rather than assuming it.

## The construction sites, since this repo has a history there

`ScheduleEntry` gained six parameters, and every production instance goes through one `fromChain()` factory, called at two sites that pass identical trailing arguments. That is the shape that prevents the bug we hit on #117, where a value object was constructed at three sites and one of them was missing an argument. Worth saying it was checked and is fine.

The reading is static throughout: no `Schedule::events()`, no booting, so a scan does not start executing registration closures to find out what is scheduled.

## One mutation on your list does not discriminate

`timezone` back in `FREQUENCY_METHODS` is listed among the mutations run and killed. It kills nothing: the whole suite stays at `426 passed`, and more to the point the behaviour does not change either. Same kernel, same output both ways:

```
with the mutation      frequency=daily   timezone=Europe/Warsaw
without it             frequency=daily   timezone=Europe/Warsaw
```

The scope fix is what does the work, and the list entry is redundant for this shape. Not a defect, and I would not remove the entry. But the mutation lists in these descriptions are load-bearing for how much a reader trusts the rest, so it is worth dropping the ones that turn out inert rather than leaving them in.

## Three stylesheets ship, two of them unreferenced

```
index-C1PztlR0.css   not referenced
index-CaA-0w68.css   not referenced
index-DIvGEsWk.css   not referenced
index-Dy8YM33d.css   referenced
index-CvLmgcc2.js    referenced
```

Three CSS commits, three intermediate stylesheets left behind. The entry chunk is handled correctly: the old one is gone and there is exactly one. Same `git rm` as #120, on the stylesheets this time.

## Everything else

The committed bundle reproduces: `npm ci && npm run build` leaves `git status` empty. `426 passed (1167 assertions)`, PHPStan clean, Pint clean, CI green across 15 jobs with per-job results. Output gate ids move only by the new schedule nodes and their edges, with null arms identical on all three applications.

## Verdict

Approving. Four defects found by reading, each confirmed with a probe before being fixed, is the right order to work in, and the one about the fixture that "would have passed any test written against it" is the kind of thing most people leave unwritten.

mrmarchone:
Conflicts @webard

PR #123, opened 2026-09-01
Feat: show which deferred providers exist, what boots them, and which never will

webard:
Deferred providers were invisible. A provider that never loads, or that loads and then still fails, looked exactly like one that works.

## What the graph now says

A deferred provider becomes a node, and every service it promises draws an edge to it — **resolving this boots that**:

Eager providers are deliberately left off. They already appear wherever a binding was wired to them, and adding the rest would scatter isolated nodes across every tab.

## What is *not* answerable, and why

The question this started from was "show me when we boot a deferred provider". That is not answerable by this tool and the feature does not pretend otherwise.

Which manifest keys a *given request* resolves is a property of one execution — it depends on branches taken, config values, middleware actually run, and lazily-touched facades. This is a static AST scanner that never boots the application. So the graph states the **cause**, which is entirely in the source: *resolving this service would register that provider*. It never claims an occurrence.

## Established from the framework source, not from memory

Read out of `vendor/laravel/framework` (v13.29.0) before any of this was written:

- **Deferral has exactly one definition.** `ServiceProvider::isDeferred()` is `return $this instanceof DeferrableProvider;` — nothing else.
- **The legacy `protected $defer = true;` is dead.** Grepping the whole framework for `$this->defer` / `protected $defer` returns nothing. A provider using it and *not* implementing the interface is registered eagerly on every request, silently, while its author believes the opposite.
- **`provides()` is the entire trigger set, not documentation.** `ProviderRepository::compileManifest()` writes `$manifest['deferred'][$service] = $provider` for each string it returns; `Application::make()` matches against that map. An empty `provides()` with no `when()` writes **no** manifest entry at all, so the provider is in neither the eager list nor the deferred map.

## Three defects that follow from that

| flag | what it means |
|---|---|
| `never-boots` | empty `provides()` and no `when()` — the provider never loads, at all |
| `unbacked-provides` | promises a service it does not bind — boots on resolution and then still fails |
| `legacy-defer` | `$defer` without the interface — eager, contrary to intent |

## Two static limits, both deliberate

Deferral inherited from a **parent** provider class is not followed — only `implements` on the class itself. That can only under-report, never mis-flag, since every defect check requires `deferred === true`.

A `provides()` that is computed, returns from inside a closure, or binds in a loop is reported as **unreadable**, and every defect check goes silent on it rather than guessing.

## About the screenshots

They are from this PR's own fixture, not from a real application, and that is the honest way to show this feature: it surfaces what a *healthy* codebase does not contain. Measured on a 60-module production application, the scanner correctly reports **0 deferred providers** — that application uses none — so there would be nothing to photograph.

The fixture holds 11 providers covering every branch: 9 deferred, one legacy-`$defer`, one plain. The analyzer reads them as such, including all three defect flags.

## Verification

`pint`, `phpstan` and the suite are green — **439 passing**, 29 new across two files. Sixteen mutations were run and all sixteen killed, including: deferral always true, `$defer` never detected, `alias()`'s second argument ignored, plain-string keys dropped, closures traversed for returns, a grouped-import head left unresolved, an absent method read as dynamic, `neverBoots` ignoring `when()`, `unbackedProvides` dropping its no-legible-bindings guard, eager providers getting nodes, boot edges for every provider, and no boot edges at all.

Three survived the first pass — each needed a fixture that actually discriminated, so `PartialProvidesServiceProvider` (half-readable `provides()`), a `$defer` alongside `DeferrableProvider`, and a legacy provider whose service really has a graph node were added. Second pass: zero survivors.

A config switch (`laravel-brain.service_providers.enabled`, default on) gates the whole walk rather than its output, so turning it off costs nothing rather than costing the same and hiding the answer.


SanderMuller (approved):
Reviewed at `33c49ee2`. Approving. Two of my three applications do use deferred providers, so I could check this against production code rather than only against the fixture, and the flags come out clean.

## The framework facts, read from the installed source

All three, against v13.29.0, the version you cite:

```
ServiceProvider::isDeferred()          return $this instanceof DeferrableProvider;
grep for $this->defer / protected $defer in the whole framework     0 hits
ProviderRepository.php:148             $manifest['deferred'][$service] = $provider;
```

Worth having checked rather than taken, because the whole design rests on the first line being the only definition of deferral.

## No false flag on 49 real deferred providers

This is the number I would want if I were merging it. Across three applications, at every scope I tried:

| | deferred providers found | never-boots | unbacked-provides | legacy-defer | unreadable |
| A | 49 | 0 | 0 | 0 | 0 |
| B | 2 | 0 | 0 | 0 | 0 |
| C | 1 | 0 | 0 | 0 | 0 |

Not one spurious defect on real code. A tool that told an engineer "this provider never loads" about a provider that loads fine would be worse than saying nothing, and it does not.

## What the graph gains, and what it does not lose

| | ids added | removed | tabs losing nodes | what was added |
|---|---:|---:|---:|---|
| A | 20 | **0** | **0** | 8 provider nodes, 12 edges |
| B | 9 | **0** | **0** | 2 provider nodes, 7 edges |
| C | 0 | 0 | 0 | nothing |

Compared per tab by id rather than by position. Null arms identical on all three.

## The default path decides how much of this anyone sees

Worth knowing, and it may apply to your own measurement. `providers.paths` defaults to `app/Providers`, and on application A only **8 of its 49** deferred providers live there. The other 41 sit beside the code they serve, in domain directories under `app/`. Pointed at `app`, the analyzer finds all 49, and its count matches a plain grep exactly.

So a reading of "0 deferred providers" on an application is a statement about `app/Providers`, not about the application. Before concluding that the 60-module one uses none, it is worth one run with a wider path. If it keeps providers beside their code the way this one does, the feature has more to show there than it appeared to.

Nothing here is this PR's doing: the path default predates it, and the analyzer honours whatever it is given. It is only that the default is what a reader will experience.

## The switch gates the walk, and costs nothing either way

```
enabled     8 provider nodes, 3,576 nodes total, 2,072 and 2,074 ms
disabled    0 provider nodes, 3,568 nodes total, 2,184 and 2,108 ms
```

Exactly eight nodes' difference, and the two arms are inside each other's noise, so the walk is not paying for itself in either direction at this size. Gating the walk rather than the output is still the right shape.

## The rest

Three of your sixteen mutations, sampled: deferral always true fails six tests, giving eager providers nodes fails `it adds no node for an eager provider that has nothing to report`, and ungating the switch fails `it does not walk the provider tree when the switch is off`. Each killed only what it should.

Chunk hygiene is right here, which is worth saying because the last two in this series were not: one entry chunk and one stylesheet, both referenced, both predecessors deleted.

`439 passed (1214 assertions)`, PHPStan clean, Pint clean, CI green across 15 jobs with per-job results. The bundle reproduces from source.

## Verdict

Approving. The section on what is *not* answerable is the best part of the description: a static scanner that says "resolving this would boot that" and refuses to say "this request boots that" is drawing the line in the right place, and saying so in the PR is what stops someone reading the edge as an occurrence later.

PR #124, opened 2026-09-01
Feat: show what each method reads, writes, forgets and locks in the cache

webard:
Cache was invisible. A method that reads through a cache, writes one, or invalidates one looked exactly like a method that does not — yet "this is cached" and "this forgets that key" are among the first things anyone needs when reasoning about stale data.

<table>
<tr>
<td></td>
<td></td>
</tr>
</table>

## What is reported

**The kind, first** — `read` / `write` / `invalidate` / `lock`. That distinction is what makes the section useful; a bare "uses cache" badge would not be worth the pixels. The method name travels alongside it, so neither is misrepresented:

- `remember` and `rememberForever` are **reads**. They write only on a miss, and what a reader needs to know is "this can be stale".
- `pull` is filed as an **invalidation**. It is the one read that also clears the key, and someone tracing *what forgets this key* has to find it.

**The key**, where it is literal or visibly constructed — `'user:' . $id` and `"user:{$id}"` both render as `user:{$id}` so the two spellings do not read differently. A key with no literal content anywhere is reported as **computed**, never as the variable holding it.

**Tags, store and a literal TTL** where declared. `flexible()` is excluded on purpose: its second argument is a `[fresh, stale]` pair, not a duration.

Recognised across the `Cache` facade (import alias resolved through the parser's use map), the `cache()` helper in all three of its forms, and any `store()` / `driver()` / `tags()` chain over either.

## Two shapes that reached the detector and produced nothing

Found by measuring against a real application rather than by reading the code:

**`Cache::lock($key)->get(fn)`** — the only way anyone writes a lock. The chain walk accepted `store`, `driver` and `tags` as intermediate hops and rejected everything else, so a lock that was *used* was dropped whole. Measured on a 60-module application: **16 `Cache::lock` calls yielded 0 lock operations**. A lock now travels with the chain and wins over the verb applied to it — the key belongs to the lock, not to the `get` that runs inside it.

**`Cache::memo()->get($key)`** hands back a memoising repository, so the operation is whatever is called on it. The same guard made every memoised read invisible; that application writes 22 of them.

The guard that matters is untouched: a `lock()` on something that is not the cache is still refused, and there is a test for exactly that.

## A test that passed for the wrong reason

The three tests written for those two fixes were green while the rules they tested were deleted. The mutation that removed the lock rule killed nothing — and only after confirming the mutation had actually applied, rather than silently failing to match, was the cause clear:

the detector's own test helper walks the whole statement and finds the inner `Cache::lock(...)` by itself, while the real pipeline hands the detector **only the statement's outermost call**. That is precisely the gap that let the suite stay green while the product reported zero locks. Those tests now run through `FlowExtractor`, the path the pipeline uses, and removing either rule turns them red.

## Where it is surfaced, and where it deliberately is not

A **panel section** in the Info tab and a **badge on the flow-chart step** — and deliberately **no canvas badge**. The canvas already stacks the N+1 and security badges at one anchor with x-offset juggling, and both earn that space by flagging something *wrong*. "This is cached" is neutral context you want once you are already looking at the node, and it needs room for the key, tags, store and TTL that make it useful.

Scope is the node's **own body**, unlike `dbQueries` which follow the call chain. Hoisting a service's `Cache::forget()` onto every action above it would put one key on a dozen panels and lose the method actually responsible.

Not covered on purpose: an injected `$this->cache` repository. Nothing in the source proves that property holds a cache rather than anything else with a `get()`, and one guessed row would undo the point of the section.

## Verification

`pint`, `phpstan` and the suite are green — **482 passing**. Twenty mutations were run and all killed, across the detector (`forget` reclassified as a read, `pull` dropped, `remember`'s TTL read from the wrong argument, a computed key printed instead of labelled, the chain walk stopped before the root, the facade alias no longer resolved through the use map, the lock hop removed, `memo` no longer passing through), `FlowExtractor` (the cache step no longer re-typed, the payload lost when a callback body is descended into), `GraphBuilder` (no descent into branches, no de-duplication, the key attached when empty) and the context exporter.

A config switch (`laravel-brain.cache_operations.enabled`, default on) gates the detector's construction rather than its output, so off means no scanning rather than scanning and discarding. One test asserts the flow-step count is identical either way: turning this off must stop adding facts to the chart, not change the chart.


SanderMuller (approved):
Reviewed at `5874f1a2`. I ran every shape in the description through the real pipeline rather than through the detector directly, and all of them come out as documented. Approving, with one thing to remove before merge.

## `mutate3.py` is committed at the repository root

99 lines of your mutation harness, added in `4d59cbfa` and never taken out again. Pint only lints PHP so nothing flagged it, and there is no `.gitattributes` in this repository, so everything in the tree goes into the composer dist: a Python script would ship to every install of a PHP package.

One `git rm` and it is gone. Worth adding `*.py` to `.gitignore` at the same time, since this is the second harness in two weeks to reach a branch by accident.

## Every documented shape, through the pipeline that ships

A fixture project with twelve cache shapes, analysed end to end:

```
lock       lock       key=job:1        literal        bare Cache::lock
lock       lock       key=job:2        literal        Cache::lock(...)->get(fn)
lock       lock       key=job:3        literal        Cache::lock(...)->block(5, fn)
read       get        key=user:1       literal        Cache::memo()->get(...)
read       remember   key=user:{$id}   constructed    'user:' . $id
read       get        key=user:{$id}   constructed    "user:{$id}"
invalidate forget     key=user:1       literal
invalidate pull       key=user:2       literal
read       get        key=user:3       literal        cache()->get(...)
read       get        key=user:4       literal        store('redis')->tags(['a'])
read       get        key=             computed       Cache::get($someVar)
                                                      $this->mutex->lock(...)  → nothing
```

Eleven of twelve, and the twelfth is the one that should be refused. The two spellings of a built key really do render identically, a computed key is labelled rather than printed, and the lock keeps its own key rather than handing it to the verb applied to it.

## The lock fix, measured on real applications

Your account is that 16 `Cache::lock` calls yielded 0 lock operations. On mine the before state was not zero, so it is worth saying what the fix actually moves here. Comparing `4d59cbfa` against the head:

| | lock operations before | after |
|---|---:|---:|
| A | 4 | **8** |
| B | 13 | **33** |
| C | 4 | 4 |

A bare `Cache::lock($key);` was already detected; what the third commit adds is the used form. So the fix widens rather than switches on, and the "0" is a fact about one application rather than about the rule.

`Cache::memo()` appears nowhere in any of my three, so that half is exercised only by the fixture above, where it works.

## The switch gates the work, and the chart does not move

```
fixture, on       15 flow steps, 11 cache operations
fixture, off      15 flow steps,  0
application B, on    4,099 flow steps, 26 cache operations
application B, off   4,099 flow steps,  0
```

Identical step counts either way on a real application, which is the property that matters: off stops adding facts rather than changing the chart.

## Output gate

Node and edge ids identical on all three applications, null arms identical, and the difference is only added keys (`cache`, `cacheOps`, `kind`, `method`, `key`, `keyKind`, `store`) with nothing removed and every other key balanced added-against-removed. A clean enrichment.

## Tests

Three of your twenty mutations, sampled: reclassifying `forget` as a read fails five tests, dropping `lock` from the recognised methods fails three, and disabling the `memo` pass-through fails `it sees through a memoising repository to the operation on it`.

I nearly reported that last one as uncaught, because it survives the two test files I ran first. It is caught in a third, and the mutation genuinely changes behaviour (11 operations become 10, the memoised read disappears). Both halves matter: a surviving mutation is only a finding once you have run the whole suite *and* checked the behaviour actually moves.

The section on the test that passed for the wrong reason is the best part of the description. A helper that walks the whole statement while the pipeline hands over only the outermost call is exactly the shape that keeps a suite green while the product reports zero, and moving those tests onto `FlowExtractor` is the right fix rather than adding another assertion.

`482 passed (1397 assertions)`, PHPStan clean, Pint clean, CI green across 15 jobs. One entry chunk and one stylesheet, both referenced, predecessors deleted, and the bundle reproduces from source.

## Verdict

Approving. Deliberately leaving out an injected `$this->cache` property, and saying why, is the call I would have wanted here: one guessed row in a section like this costs more than the rows it would add.

PR #125, opened 2026-09-01
Feat: show which laravel/ai agents exist, what they can call, and what sets them off

webard:
`laravel/ai` agents and tools were being scanned and then thrown away. On a 60-module application the pass found **5 agents and 17 tools** — and produced **0 edges**, leaving all 22 nodes in no tab at all. Nothing about that was visible from the summary: the scan reported the agents it had found, and the viewer showed none of them.

<table>
<tr>
<td></td>
<td></td>
</tr>
</table>

## Three separate causes, only one of them the expected one

**Call sites were found, not missed.** The analyzer reported six — a translator service, a queued job's `streamResponse()`, a scripted assistant, an intent detector. The `new SomeAgent(...)` shape already matched. They were dropped one step later: `aiCallerNodeId()` looked up an existing node and gave up when there was none, and every one of those six classes had no node on the graph — a job, several services, none of which the call-chain tracer reaches from any route. The fix is not new matching but **creating the caller node**, classified as the tracer would have, the way `addFilament()` already creates a minimal model node.

**`tools()` that returns an injected property is now reported, not passed over in silence.** The distinction is drawn at the source: a `tools()` whose every returned item is a class reference is readable; anything else — property fetch, container call, generator, spread — sets `toolsAreDynamic`, and the panel says so outright. `return [];` still means an agent with genuinely no tools, which was the distinction worth not losing.

**Tools handed to the constructor are attributed.** A class named in a `new SomeAgent(...)` argument list whose own body instantiates recognised tools is treated as supplying them, so an agent built with `tools: $provider->toolsFor($user)` reaches all 17 of that provider's tools. Those edges are labelled **`may call (supplied)`**, kept apart from declared ones because the two are known with different certainty. An agent or a tool is never treated as a provider.

## Fixing the edges did not make them visible

After all three fixes the application had 23 AI edges and zero isolated AI nodes — and **still 22 of 22 AI nodes in no tab**, because tabs are seeded from routes, commands, channels and Filament, and those caller clusters carry no seed. Wiring was never the whole problem.

So agents get a standalone **AI Agents tab**, the way models get the ERD tab, assembled node by node — agents, tools, and their direct callers — rather than by walking forward from a seed, which would drag in each caller's entire downstream subtree. On that application the tab is **28 nodes / 23 edges, and 22 of 22 AI nodes now appear in it**.

## The scan summary now reports both kinds of invisibility

An edgeless-node check alone would have reported nothing wrong here — after the edge fix the AI nodes were wired and still unreachable. So the summary reports two numbers, separately, each with its top three types:

- **Isolated nodes** — a pass that built nodes and forgot to wire them.
- **Outside tabs** — wired, but no tab seed reaches it.

The second is the one that catches this class of defect. Both are reported as baselines, not chased: on the measured application they stand at 207 and 1405, and several entries look like real gaps in other passes.

## What the panel states

Model tier and its source (`#[UseCheapestModel]` renders as *cheapest tier — provider decides*, never as a model id the code does not name), provider, implemented contracts, which methods the agent overrides, and the two tool facts kept apart: what `tools()` declares, and how many were supplied at construction. Tools reachable only through a runtime container binding or a variable stay unnameable and are reported as such rather than guessed.

## Verification

`pint` clean, `phpstan` no errors, **444 tests passing** — all with `laravel/ai` **not installed**, so the pass degrades to nothing rather than fataling when the package is absent. Fixtures cover the shapes; the package is a dev-only requirement.

Twenty-three mutations were run and killed, covering: an unreadable `tools()` reported as no-tools, a literal list wrongly called dynamic, injected tools never attributed, agents and tools treated as providers, construction arguments not read, the caller node no longer created, supplied tools sharing the declared label, the AI tab never built, the AI tab built without agents, and both new counters (one marking everything it touched, one counting nothing).

A config switch (`laravel-brain.ai.enabled`, default on) gates the analyzer's construction rather than its output, so off means no scanning rather than scanning and discarding.


SanderMuller (approved):
Reviewed at `076f2863`. One of my three applications turns out to use `laravel/ai`, so the whole thing could be measured on production code rather than on the fixture alone. Approving.

## Measured on a real application

That application has 36 files referencing `Laravel\Ai`, and the pass lands cleanly on it:

| | on `main` | on this branch |
|---|---:|---:|
| AI nodes | 0 | **13** |
| AI edges | 0 | **18** |
| isolated AI nodes | 0 | **0** |
| AI nodes in no tab | 0 of 0 | **0 of 13** |
| tabs | 370 | 371 (`ai--agents`) |

"Zero in no tab" is the number the third commit exists for, and it holds on an application that is not the one it was built against.

The full-graph difference there is 47 ids added and **none removed**: 13 AI nodes, 18 edges, and 16 caller nodes that no other pass had created, which is exactly the fix you describe. No existing tab loses a node. The other two applications are byte-identical, and the null arms were identical on all three.

## Every edge kind, on the fixture

```
may call             x3      declared tools
may call (supplied)  x2      handed to the constructor, labelled apart
delegates to         x1
prompts              x4
```

On the real application all 18 edges are `prompts`, so the supplied-tools path is exercised only by the fixture. Worth saying plainly rather than letting the fixture stand in for evidence it cannot give.

## The switch

```
on    13 AI nodes, 18 edges, 371 tabs
off    0 AI nodes,  0 edges, 370 tabs
```

Off removes the pass and its tab rather than filtering their output, which is what gating construction buys.

## Tests

Three of your twenty-three mutations, sampled, each killing only what it names:

| mutation | fails |
|---|---|
| `aiCallerNodeId()` returns null again | `it wires the controller action that prompts an agent`, `it creates the caller node when no other pass reached that class`, `it puts every AI node in the AI Agents tab` |
| supplied tools share the declared label | `it wires tools supplied at construction, labelled apart from declared` |
| an unreadable `tools()` reported as readable | `it flags a tools() body it cannot read, instead of reporting no tools`, `it says on the node that tools are decided at runtime` |

`444 passed (1248 assertions)`, PHPStan clean, Pint clean, CI green across 15 jobs. One entry chunk swapped, its predecessor deleted, and the bundle reproduces from source.

The two counters are the part I would keep even if the AI pass were dropped. An edgeless-node check would have reported nothing on the state you describe, and "wired but in no tab" is the question that actually catches it. Reporting both as baselines rather than as failures is the right call at 207 and 1405.

## One small correction

The description says the package is a dev-only requirement. It is not in `require-dev`, and it does not need to be: the fixtures are plain PHP naming `Laravel\Ai\...` classes that are never loaded, which is why the suite passes with the package absent. The substance is better than the sentence, so it is the sentence that wants changing.

## Verdict

Approving. Finding three separate causes behind one symptom, and saying which of them was the expected one, is the part that makes this reviewable. The honesty about the tab seeds ("wiring was never the whole problem") is what turned it into a real fix rather than a partial one.

Worth a note for my own account: `composer test` failed for me at first with `Undefined constant "Larastan\Larastan\LARAVEL_VERSION"`, which looked like this branch breaking the type check. It was PHPStan's result cache, shared through the system temp directory across every worktree on this machine. `phpstan clear-result-cache` and it is green. Nothing here to fix; recording it so the next person who sees it does not spend the same twenty minutes.

PR #126, opened 2026-09-01
Feat: show which methods call out over HTTP, and whether the call can hang

webard:
A method that calls an external API looked exactly like one that does not. "This talks to eBay", "this has no timeout", "this retries three times" are among the first things anyone needs when reasoning about a request that can hang, and none of them were anywhere in the graph.

<table>
<tr>
<td></td>
<td></td>
</tr>
</table>

## What is reported

**The verb and the path**, where either is literal or visibly constructed. A path assembled from a literal and a variable renders with the variable in place rather than as the variable name, so two spellings of the same call do not read differently.

**The client** — the Laravel HTTP client, Guzzle, or a curl call — because the three fail differently and the reader is usually looking for one of them specifically.

**Timeout and retry**, which is the point of the section. `no timeout` is styled as a warning and `no retry` is not: a missing timeout is usually an oversight, a missing retry is usually a decision. A literal timeout travels with its value.

**The host**, where the base URL is a literal or a config key. Where it is neither — `Http::baseUrl($this->getApiUrl())` is a method call, and nothing in the source says what it returns — the call is reported as **external** and the path as **partly computed**, never as a guessed address. Half the value of this section is that it does not invent the half it cannot see.

## The shape that reached the detector and produced almost nothing

Found by measuring against a real application, not by reading the code. A client class that builds a configured request in one method and sends it from another:

```php
$this->client->api()->get('/commerce/identity/v1/user/');
```

`api()` returns a `PendingRequest` with the base URL, auth and timeout already applied — but the send site is a method call on a property, and the detector had no way to know that property held a request. Measured on a 60-module application: **7 nodes and 13 calls, against 50 files that contain HTTP calls**; nearly every real client in that codebase writes them this way.

`PendingRequestAnalyzer` indexes methods **declared** to return `Illuminate\Http\Client\PendingRequest` and joins the send site to the builder, so the settings applied where the request is built travel to where it is sent.

**Detection can be sound here; attribution cannot.** The index is keyed on method name, so two classes with a `api()` method that both return a `PendingRequest` are indistinguishable at this level, and the settings of one can be attributed to the other. That limitation is stated in the class docblock, in `declaredBuilderSettings()`, in `docs/how-it-works.md`, and pinned by two tests — one of which is the false-positive case, asserting the behaviour rather than pretending it does not exist.

## Where it is surfaced

A **corner marker on the canvas node** (`external`, or the host when known), a **chip in the panel header** counting the calls, and a **panel section** carrying verb, path, client and the timeout and retry facts. The marker is deliberately quiet: the canvas already reserves loud treatment — a red border and a badge — for things that are *wrong*, and an outgoing call is not wrong. It is context you want once you are already looking at the node.

## Verification

`pint` clean, `phpstan` no errors, **471 tests passing**. Eighteen mutations were run and killed across the extractor, the builder index and the graph builder, including one that initially survived in the loop branch and is now covered by an instance-wrapper test.

A config switch (`laravel-brain.outgoing_http.enabled`, default on) gates the detector's construction rather than its output, so off means no scanning rather than scanning and discarding.


SanderMuller (approved):
Reviewed at `0c590b58`. Approving. The mechanism does what you say, and the honest limit is worth stating alongside it.

## The builder fix, before and after

A fixture with the shape the last commit exists for: a client class whose `api(): PendingRequest` sets a base URL, a timeout and a retry, and a controller that sends from it.

```
                                            first commit   head
$this->client->api()->get('/…/user/')       not detected   detected, host api.example.com
Http::timeout(2)->get('https://direct…')    detected       detected
Http::baseUrl(config('services.foo.url'))   detected       detected
Http::baseUrl($this->client->api())         no host        no host
```

The settings really do travel from where the request is built to where it is sent, and the last line is the one I most wanted to see: a base URL the source cannot resolve produces no host rather than a plausible wrong one.

## What it finds on real applications, and why that number is small

Three applications, node and edge counts unchanged, only http keys added, nothing removed, null arms identical:

| | files containing `Http::` | files declaring a `PendingRequest` builder | nodes carrying calls |
|---|---:|---:|---:|
| A | 2 (plus 20 Guzzle) | 1 | 3 |
| B | 20 | 13 | 3 |
| C | 21 | 22 | 3 |

Detection is identical at the first commit and at the head on all three, so the later two commits buy nothing *there*. The reason is the tracer rather than the detector: on C, of the 22 files declaring a builder, only 6 have any node in the graph at all. The rest are never reached from a route or a command, so there is no node to hang a call on.

That is the same bound as the deferred-provider work, and it is worth saying in the description rather than leaving a reader to infer that 21 files with `Http::` should have produced 21 marked nodes. Your own "7 nodes and 13 calls against 50 files" is the same ratio seen from the other side.

## The switch

```
fixture      on: 4 calls    off: 0
application  on: 3 calls    off: 0
```

Off removes the pass rather than filtering it, as claimed.

## Tests

Two of your eighteen mutations, sampled against the whole suite:

- the builder index always returns empty → four tests fail, including `it finds the call a client class builds and a service sends, end to end` and the config-switch test
- every client classified as the Laravel client → `it reports a call on a client constructed in the same method` and the deduplication test

Each fails only what it names.

`471 passed (1281 assertions)`, PHPStan clean, Pint clean, CI green across 15 jobs. One entry chunk and one stylesheet, both referenced, predecessors deleted, and the bundle reproduces from source.

## What I checked and found right

The claim I expected to be shakiest is the one about attribution, and you have already written it down where it belongs: keying the index on method name means two classes with an `api()` method are indistinguishable, and the settings of one can land on the other. Saying so in the class docblock, in the helper, in `docs/how-it-works.md` and in a test that asserts the false positive rather than hiding it is the right handling. A tool that silently attributed one client's timeout to another would be worse than one that says it might.

`no timeout` styled as a warning and `no retry` not is the same instinct: a missing retry is usually a decision, and treating the two alike would train people past both.

## Verdict

Approving. Reporting a call as `external` with a `partly computed` path, rather than guessing the host, is what makes the rest of the section trustworthy.

mrmarchone:
Conflicts @webard

PR #127, opened 2026-09-01
Feat: show the name each model goes by in a morph map

webard:
A model's morph alias is the name it goes by everywhere it is stored polymorphically — in `*_type` columns, in `morphMap` lookups, in every `whereMorphedTo` query. The graph knew the class and the table but not that name, which is the one an operator needs when they are staring at a `taggable_type` value and trying to work out which model it is.

<table>
<tr>
<td></td>
<td></td>
</tr>
</table>

## What is read

`Relation::morphMap([...])` and `Relation::enforceMorphMap([...])` wherever they are registered — in this codebase that is a service provider, but nothing about the reading is provider-specific. The map is resolved statically, so an alias registered in a package the application never boots is still found.

The alias is reported on the model's schema section, alongside the table and primary key it already carries.

## The verdict, not the global flag

`enforceMorphMap()` (and `requireMorphMap()`) removes the fallback to the class name: the **first** `getMorphClass()` call on an unmapped model throws `ClassMorphViolationException`. So a missing alias is only worth reporting where enforcement is on — everywhere else the class name works fine and a warning would be noise.

The model node therefore carries the verdict — *this model, under this application's rules, is unmapped* — rather than the global "is enforcement on" flag. Otherwise every one of the 181 model nodes would repeat the same fact about the application, and none of them would say anything about themselves. Measured on a 60-module application that calls `requireMorphMap()`: **133 models mapped, 46 flagged**.

## One false positive, found by measuring rather than reading

The first version flagged **48**, and two of them were abstract base classes. An abstract class cannot be instantiated, so no `getMorphClass()` call can ever reach it and a map that omits it is correct rather than incomplete — the tool was reporting a defect those classes can never cause.

Abstractness was not available at the decision site: the analysis visitor never handled `Stmt\Class_` at all, so it was reading everything about the class except whether it could exist. It now does, and the flag requires a concrete class. A fixture with an abstract model pins it; removing the condition turns that test red.

## Verification

`pint` clean, `phpstan` no errors, **423 tests passing**. The fix above was verified red-green — the new test fails against the previous condition and passes against the current one — and re-measured end to end on the real application, where the count moved 48 → 46 and both abstract classes stopped being flagged.

A config switch (`laravel-brain.morph_map.enabled`, default on) gates the lookup's construction rather than its output, so off means the map is never read rather than read and discarded.


SanderMuller (approved):
Reviewed at `9323452f`. The reader is right and I checked every branch of it directly. One thing in the description says the opposite of what the code does, and that is the only finding.

## The description and the implementation disagree about the mechanism

The description says:

> The map is resolved statically, so an alias registered in a package the application never boots is still found.

`MorphMap::fromApplication()` calls `Relation::morphMap()` and `Relation::requiresMorphMap()` on the running framework. There is no parsing anywhere in the class, its docblock argues explicitly *against* parsing providers ("Parsing providers for `morphMap([...])` literals sees a strict subset of that"), and one of its tests is called `it reads the aliases the running application registered`.

Run against the class:

```
nothing registered       aliasFor(Parcel) = NULL     enforced = false
registered at runtime    aliasFor(Parcel) = 'parcel' enforced = false
plus requireMorphMap()   aliasFor(Parcel) = 'parcel' enforced = true
switch off               aliasFor(Parcel) = NULL     enforced = false
```

A package that never boots registers nothing, so its alias is exactly what this cannot find. The code has the better of the argument and I would not change it. The sentence is what wants changing, because someone reading the description would expect a guarantee the feature deliberately does not make.

## Everything else in the reader, checked branch by branch

The duplicate-alias case is the subtle one and it is right: a class listed under two keys resolves to the **first**, which is what `array_search()` inside `Model::getMorphClass()` answers, so the alias shown is the one the database actually holds.

```
map: ['second' => Twice::class, 'first' => Twice::class]
aliasFor(Twice) = 'second'      array_search would answer 'second'
```

The switch returns an empty map without asking the framework, so off means the scanner never reads rather than reads and discards.

## The abstract-class fix

No application I measure with enforces a morph map, so the flag cannot fire on any of them and I cannot reproduce your 48 → 46 on real code. The mutation is the evidence instead: replacing `&& ! $visitor->isAbstract` with `&& true` fails `it does not flag an abstract model an enforced map leaves out`, and nothing else. Removing the framework read entirely fails three, including the two that name the runtime behaviour.

Both run against the whole suite rather than the file whose name matches, which is how I have been catching my own false "nothing covers this" lately.

## Output gate

Node and edge ids identical on all three applications, null arms identical, and the difference is two added keys, `morphAlias` and `morphAliasMissing`, with nothing removed and everything else balanced.

Every alias comes out `null` in my runs, and that is the feature working rather than failing: my harness analyses a project without booting its providers, so `Relation::morphMap()` is empty, exactly as the class docblock says it will be. One of the three applications does register a map in a provider, and it still reads empty here for that reason. A real scan runs from inside the booted application through `brain:scan`, where it will not.

`423 passed (1149 assertions)`, PHPStan clean, Pint clean, CI green across 15 jobs. One entry chunk and one stylesheet, both referenced, predecessors deleted, bundle reproduces from source.

## Verdict

Approving. Carrying the per-model verdict instead of the global flag is the right call and the reasoning for it is the part I would keep: a fact repeated on 181 nodes is not a fact about any of them. Fix the sentence about static resolution and the description matches the thing you built.

PR #128, opened 2026-09-01
Feat: support the OPTIONS and QUERY HTTP methods

webard:
Two HTTP verbs the viewer could not name. One of them was already in the data.

<table>
<tr>
<td></td>
</tr>
</table>

## OPTIONS was extracted from the start and never once displayed

`RouteAnalyzer` has recognised `Route::options()` in its `HTTP_METHODS` list since the beginning, and the live-router path yields OPTIONS from `$route->methods()` like any other verb. The gap was on the other side: membership in the frontend's `METHOD_COLORS` is what makes a verb visible at all — `splitLabel`, `routeSegments` and the chip filter each ask whether a label's first word is in that map before treating it as a method.

A verb the analyzer emits but the map omits therefore renders its **entire** label as the URI. An `OPTIONS /health` route displayed as a path literally called "OPTIONS /health", sorted under that spelling, and could not be filtered. Measured on a 60-module application: one such route, misrendered that way for as long as the viewer has existed.

## QUERY, without pretending it exists yet

QUERY (RFC 9110bis — safe and idempotent, carries its query in the request body) is in the verb list, but nothing here guesses it into a route. `Route::query()` and the `Router::$verbs` entry were merged in **laravel/framework#60655**, which went to `master` (the 14.x line) and not to 13.x. On 9 through 13 the only spelling is `Route::match(['query'], …)`.

Hard-coding either verb list would be wrong on half the support matrix — it would drop a real QUERY endpoint on 14.x or invent one on 13.x. The shared list deliberately never includes QUERY, so the method is recognised where the framework has it and absent where it does not.

## The two new colours were measured, not eyeballed

Against the five existing chips the tightest pair is POST/PATCH at **ΔE2000 19.3**, so that is the bar a new colour has to clear. Cyan (OPTIONS) sits at ΔE 22.4 from its nearest neighbour, pink (QUERY) at 18.6. Both hold up in either theme: contrast is 10.5 and 7.2 against the dark panel and 1.8 and 2.7 against the light one, inside the 1.7–2.8 band the existing five already occupy — and a toggled chip additionally paints a 16% tint and a 55% border of its own colour.

The chip row now wraps. At the 280px minimum sidebar width seven chips leave roughly 32px each while "OPTIONS" needs about 42px at 10px mono, so a fixed row would have clipped it to "OPTIO". `flex: 1 1 auto` lets a short row still fill the width and a long one spill onto a second line.

## Verification

`pint` clean, `phpstan` no errors, **420 tests passing**, and confirmed end to end on a real application: the seven chips render, and the previously misrendered OPTIONS route now carries a method badge and a path that can be filtered.


SanderMuller (approved):
Reviewed at `86d0019`. Suite green in a fresh worktree: **420 passed (1167 assertions)**, PHPStan **No errors**, Pint clean.

This is the rare PR where the description's central claim is the one worth testing and it holds all the way through. `Route::any()` really did produce a single `ANY /uri` tab that the sidebar could not colour or filter, while live-router mode had been emitting one route per verb since forever. Both readers now agree.

## Where the review went

**The load-bearing claim — source mode and live-router mode produce the same tabs.** `discoverFromRouter()` already loops `$route->methods()`, skips HEAD and builds `tabGroup` as `VERB . ' ' . $uri`; `resolveRouteVerbs()` now does the same three things on the static path. I ran both readers over the same registrations (`match(['QUERY'], …)` and `options()` against a real `Router`) and got the same verb sets. Your own test does this too.

**Output gate on three applications.** Null arm first (base against base, same corpora, same session) — identical on all three, so the differences below are the branch's. Every one of them is explained:

| | routes before | after | what moved |
|---|---:|---:|---|
| A | 447 | 482 | 7 `ANY` tabs → 42 verb tabs |
| B | 341 | 347 | 3 `Route::match()` registrations that were previously read as nothing |
| C | 359 | 434 | 15 `ANY` tabs → 90 verb tabs |

I checked the replacement is exact rather than approximately right: for every `any-…` subgraph key that disappeared, all six `get-…`/`post-…`/`put-…`/`patch-…`/`delete-…`/`options-…` keys exist on the branch. No tab was lost, none changed identity. On B the six new tabs also drag in 9 service, 1 controller, 1 middleware, 1 view and 1 action node that nothing could reach before — those `match()` routes were invisible, not merely mislabelled.

**Mutation checks**, whole suite each time, restore verified:

- keep HEAD in the `match()` verb list → 1 failed
- hard-code `anyRouteVerbs()` to a list containing QUERY → 1 failed
- read the action from a fixed `args[1]` instead of `$uriArg + 1` → 2 failed

**The bundle reproduces.** `npm ci && npm run build` leaves `git status` empty, and the branch adds exactly one entry chunk against its merge-base — a rename, so the old one is gone rather than accumulating. Thank you for that; it is not the default here.

**The sidebar, run rather than read.** I rendered `LeftSidebar` under jsdom with three tabs (`GET /users`, `OPTIONS /preflight`, `QUERY /search`) on both sides. On `main` the chip row is `GET POST PUT PATCH DELETE`, the mode tab says `Routes 3`, and the tree shows one route: the other two fall into a collapsed `Other 2` group, because `routeSegments()` returns null for a first word it does not recognise. So the count and the list disagreed, which is a nastier symptom than "no chip". On the branch all seven chips render, all three routes appear in the tree, and switching OPTIONS off takes `/preflight` out and leaves `/search` and `/users` in.

**The CSRF comment's citation checks out.** `PreventRequestForgery::isReading()` on the installed framework is `in_array($request->method(), ['HEAD', 'GET', 'OPTIONS'])` — QUERY is not a read there, so a token is required and offering the checkbox is right. `Router::$verbs` here has no QUERY, which is what makes your `anyRouteVerbs()` assertion a real one rather than a tautology.

## Three things worth knowing, none of them blocking

**1. `query` and `match` widen an existing false-positive hole in the visitor.** The `MethodCall` branch of `enterNode()` has no receiver guard — any method call whose name is in the verb list is treated as a route registration, wherever it sits. Traversal continues into route closures, so:

```php
Route::get('/users', function (Request $request) {
    $sort = $request->query('sort', 'name');
});
```

`main` emits one route from that file. This branch emits two — the second is `QUERY /sort`.

The hole is not yours: `main` does the same for `$request->get('page')` and `$request->options('flags')`, which I confirmed on both arms. But `$request->query('literal')` is more idiomatic than either, so the branch makes an existing defect easier to hit. It fires zero times across the three applications I gate on (no `->query(` or `->match(` anywhere in their route files), which is why this is a note and not a request. A receiver guard on the `MethodCall` branch would close it for all seven verbs at once, and belongs in its own PR.

**2. The `$verbs === []` guard in `handleHttpRoute()` is a short-circuit, not a safety net.** I deleted it and the suite stayed green — including `skips a Route::match() whose verb list is built at runtime`. That is not a coverage gap; the mutation genuinely changes nothing, because the `foreach ($verbs as $verb)` below already emits nothing for an empty list. The test is correct about the behaviour, it just pins the loop rather than the guard. Keep the guard, it saves the work; I mention it only so nobody later reads a green suite as proof the early return is load-bearing.

**3. Multi-verb expansion multiplies an existing duplicate-edge problem.** Some controller-to-interface and controller-to-action edges are already emitted once per route reaching the controller, so a controller behind an `ANY` route goes from one copy to six. In one tab on A the node set is unchanged and the edge list goes 121 → 171. Across the whole graph on A duplicate edge copies go 2,721 → 2,866. The duplication predates this PR and follows from a route becoming six routes, so nothing here is wrong; it is a good argument for deduplicating per subgraph in a later change.

Also worth being deliberate about: on C, an application with 15 `Route::any()` registrations, the sidebar goes from 424 tabs to 499. Six rows where there was one is the price of matching live-router mode, and I think it is the right trade — but it is the change users will notice first, and the description does not mention it.

## Screenshots, please

The chip row now wraps and carries two new colours. Could you attach a before/after of the sidebar at the 280px minimum width, with all seven chips on, and a second pair with OPTIONS switched off so the filtering is visible? If you have no capture setup, these two skills do it: [eye-verification](https://github.com/SanderMuller/boost-skills/blob/main/resources/boost/skills/eye-verification/SKILL.md) and [frontend-quality](https://github.com/SanderMuller/boost-skills/blob/main/resources/boost/skills/frontend-quality/SKILL.md).

The CIELAB numbers in the `METHOD_COLORS` comment are the right way to justify two new colours and I did not re-derive them; a picture would settle the wrap behaviour, which no number in the diff describes.

## Verdict

Approving. One list instead of three, both readers in agreement, the constant wired through to the stress endpoint's allow-list rather than copied, and tests that fail when you break them.

PR #129, opened 2026-09-01
Feat: recognise single-purpose action classes as their own kind

webard:
A single-purpose action class — one class, one public entry method, one deliberate unit of work — was rendered as a plain `service`, the kind the pipeline falls back to when nothing else recognises a class. The pattern is one of the more common ways a Laravel application organises its work, and the graph had no name for it.

<table>
<tr>
<td></td>
<td></td>
</tr>
</table>

## The directory is the signal, not the name

The pattern has no naming convention worth matching. The classes are called `CreateOrder`, `PublishPost`, `ChargeCustomer` — verbs, with nothing in the identifier to key on. A `*Action` suffix test would miss every one of those *and* claim every Filament table action in a Filament application.

What the pattern does have is a placement convention: the classes are gathered under an `Actions/` directory, which is what `lorisleiva/laravel-actions`, `spatie/laravel-queueable-action` and the hand-rolled variant all write, and what makes the group legible in the first place. So the directory is the signal — exactly as it already is for listeners and observers.

Classification is deliberately conservative: it only ever fires against a class the rest of the pipeline already gave up on and called a plain `service`. A job, listener, model, event, facade or Form Request that happens to sit under `Actions/` keeps the kind that recognised it, because that kind says more.

## Two names that were the same word

Controller actions were already called "actions" in the viewer. Rather than overload the word, the two are now distinguished in the legend and the filter panel: **Controller actions** for a controller method reached by a route, **Actions** for these classes.

## The entry method, and when a caller goes around it

The node carries `entryMethod` — the class's single public entry point, one of `__invoke`, `handle` or `execute` — alongside `method`, the one the incoming hop actually called. Normally the same string. When they differ, a caller has reached past the entry point into a helper, which is the thing the single-entry-point pattern exists to prevent, and the two fields show it rather than hiding it behind one label.

Where a class declares two of those methods publicly, nothing is claimed: it is not the single-entry-point unit this describes, and naming one of them as *the* entry point would be a guess presented as a fact. It stays an action class by placement; the graph just does not say where to start reading.

## A gap found in review, not in the tests

Verified against a real modular monolith, the feature classified **nothing** — 254 action classes present, several reached by the tracer, zero recognised.

The containment check compared unresolved paths. Every other consumer of that check is fed by a directory walk, but this one is handed a file the **call-chain tracer** produced, and the tracer resolves classes through Composer's autoloader — so in an application built on path repositories, a module's files arrive spelled through the symlink under `vendor/` while the configured root names the real directory. One directory, two spellings; a prefix test matches neither.

Both sides are now canonicalised. Resolving only the file would have been worse than resolving neither, since `realpath()` also collapses repeated separators — a root spelled `tests//fixtures/project` would have stopped matching its own files. `SourceDirectories::contains()` itself is untouched: its other two callers already hold real paths and pay nothing for a resolution they do not need.

After the fix, on that same application: **6 of the 6 action classes the tracer reaches are classified, and the `service` count drops by exactly those 6.**

## Verification

`pint` clean, `phpstan` no errors, **431 tests passing**. The fix above was verified red-green with the mutation confirmed applied, and its test cleans up the symlink in a `finally` — the link grafts a second copy of a package into the shared fixture, and leaving it behind makes every later test that scans the project count it twice.

A config switch (`laravel-brain.actions.enabled`, default on) gates the classifier's construction, and `laravel-brain.actions.paths` names the roots — globs are expanded, so a modular monolith points it at a wildcard segment.


SanderMuller (approved):
Reviewed at `64ef266`. `composer test` in a fresh worktree: **431 passed (1166 assertions)**, PHPStan **No errors**, Pint clean — the count in your description, on the nose. All 15 CI jobs green.

This is a new node kind added without disturbing anything else, and the gate says so more clearly than I expected it to.

## The output gate

Null arm first — base against base, three applications, identical. Then the branch, and the diff is the smallest shape a new kind can have:

| | nodes | edges | tabs | flipped `service` → `action_class` | `calls` → `runs` |
| A | 3,568 → 3,568 | 8,686 → 8,686 | 537 → 537 | 1 | 2 |
| B | 2,889 → 2,889 | 5,218 → 5,218 | 370 → 370 | 205 | 258 |
| C | 2,573 → 2,573 | 5,286 → 5,286 | 424 → 424 | 39 | 56 |

Not one node added or removed, not one tab added, removed or otherwise changed. The only fields that move on any node are `type`, `data.subtype` and `data.entryMethod`; the only edge change is the label. Every flipped node's label is byte-identical to before.

I checked the reclassification does not drop anything the node was carrying: across the 245 flipped nodes, zero `hasN1`, `fatMethod`, `validationRules`, `complexity`, `loc` or `methods` values were lost, including all 13 that carried `hasN1`. The tracer's own `'service'` lists in `MethodTracer` and `QueryTracer` are trace types rather than graph types, so the flip happens after they have had their say — that is the reasoning, and the 13 preserved `hasN1` marks are the measurement.

**Every flipped file sits under `app/Actions/`.** Two of the three applications also carry `app/Filament/Actions` and one carries `app/Nova/Actions`; nothing from either was claimed, which is the Filament false positive you named in the description not happening. `SourceDirectories::contains()` appends the trailing separator before comparing, so `app/Actions` does not swallow `app/Actionsomething` either.

## Mutations

Seven, whole suite each time, anchors asserted before running, restores confirmed with `git status --porcelain`. All seven caught:

| mutation | result |
|---|---|
| drop the `$traceType !== 'service'` early return | 12 failed |
| check `action_class` before `validation_request` | 5 failed |
| name the first entry method when two are declared | 1 failed |
| drop the `realpath()` canonicalisation | 1 failed |
| ignore the `enabled` flag | 3 failed |
| treat an empty `paths` array as keep-the-default | 3 failed |
| let non-public methods count as entry points | 2 failed |

The first one is the conservative claim, and 12 failures is the fixture doing its job.

## Probes

**The symlinked module, on a real symlink rather than the fixture.** I built a throwaway project with `app-modules/billing/src/Actions/ChargeCard.php` reached through `vendor/acme/billing`, pointed `actions.paths` at `app-modules/*/src/Actions`, and ran the analyzer:

```
with realpath (as shipped):  action_class  billing_actions_chargecard  entryMethod=handle
realpath removed:            service       billing_actions_chargecard  entryMethod=-
```

So the fourth commit is load-bearing on the shape it was written for, not only on the test that pins it.

**Kinds that are not classes.** An `interface`, an `enum` and a `trait` placed under `app/Actions/` alongside a real action class: the interface stays `interface`, the enum stays `enum`, and only the concrete class becomes `action_class`. That was the one shape the fixture project does not include and I could not talk myself into assuming.

**Both config gates.** `actions.enabled => false` and `actions.paths => []` each put the same class back to `service`; a `paths` list containing a glob still classifies it. Same fixture, same class, three configurations.

**The viewer.** `npm ci && npm run build` leaves `git status` empty, and the branch swaps exactly one entry chunk. Rendering the real components: the legend shows `Controller action` and `Action` as two rows and carries `#84cc16`; the filter panel shows `Controller actions` and `Actions` with their own counts. All four colour maps carry the kind, and a sweep for type lists mentioning `validation_request` without `action_class` comes back empty.

## Three notes, none blocking

**1. `method` differs from `entryMethod` far more often than the reading in the comment allows.** The docblock says a difference means "a caller reaching past `handle()` into a helper, exactly the thing the pattern is supposed to prevent". Measured:

- 176 of the 245 `action_class` nodes have `method !== entryMethod` — 72%.
- 32 of those are `__construct`, which is instantiation, not going around anything.
- Of the 144 remaining, **141 are reached only from the same class** — the action's own entry method calling its own public helper, which the graph gives its own hop node.
- **3** across all three applications are an outside node reaching past the entry point. That is the case the comment describes, and it is 1.2% of them.

Nothing misleads anyone today: `entryMethod` is not read anywhere in the viewer yet. I raise it because it decides how the field should be rendered when something does read it. As it stands, a warning on "these two differ" would fire on 72% of action-class nodes and be right about 3.

**2. What the classifier enforces is placement, and the opening line of the description promises more than that.** Seven of the 245 flipped classes declare none of `__invoke`, `handle` or `execute` publicly — one of them exposes three unrelated public methods and nothing else. They are helpers that happen to live in `Actions/`. The classifier calls them action classes and honestly declines to name an entry method, which is the right behaviour and exactly what your `ActionClasses` docblock says ("It stays an action class by placement"). It is only the description's first sentence — "one class, one public entry method" — that reads as a test being applied rather than a pattern being described. Worth a clause, since it is the first thing a reader meets.

**3. This collides with #128 on the bundle.** Both branches delete `resources/assets/assets/index-B5C0iocj.js` and rewrite line 11 of `index.blade.php` to a different hash. Whichever merges second gets a conflict there plus a chunk built from the wrong source. The resolution is `npm run build` and a commit of what it writes, never picking one of the two hashes by hand — the hash is a content address, and after a merge neither of them is the content.

One smaller thing, inherited rather than introduced: `SourceDirectories::resolve()` memoizes on a static keyed by root and patterns, so a glob is expanded once per process. A module added during a long watch session will not be picked up until the process restarts. That is already true of the source paths, and it costs nothing to leave alone.

## Verdict

Approving. The kind is additive, the gate proves it, the conservative ordering is pinned by a mutation that fails twelve tests when you break it, the config gate is a real gate in both directions, and the symlink fix bites on an actual symlink. Screenshots were in the description already, which is why this review has no request at the end of it.

mrmarchone:
Conflicts @webard

PR #130, opened 2026-09-01
Feat: draw the jobs that go out together, in the order they run

webard:
> **Merge #117 first.** This branch is built on  and rebased onto its current head, so the diff below **includes #117's three commits** — GitHub cannot base a PR on a branch that lives in the fork. Review #117 there; once it merges, the diff here collapses to the chain and batch work alone. The commits that belong to this PR are the last four: reading chains and batches, drawing a chain as a sequence, the config switch, and two fixes described below.

A transaction was the only thing the canvas could draw a boundary around. Jobs that go out **together** — a chain that runs in order, a batch that runs at once — were drawn as unrelated nodes that happen to sit near each other, which is the one thing they are not.

<table>
<tr>
<td></td>
<td></td>
</tr>
<tr>
<td align="center"><sub>Four region kinds on one route</sub></td>
<td align="center"><sub>The label defect this found, before the fix</sub></td>
</tr>
</table>

## A region is a kind, not a transaction drawn differently

The machinery #117 built for transactions is generalised rather than copied: a region has a **kind**, and `transaction`, `rollback`, `chain` and `batch` are four of them. The compensation path becomes a region in its own right instead of a transaction wearing a different colour — it runs with the transaction already gone, which is a different fact about the work, not a different shade of the same one.

Each kind gets its own switch in the legend, because a canvas with three chains on it is a different picture from one with three transactions.

## What is read

Five dispatch forms: `Bus::chain([...])`, `Bus::batch([...])`, `dispatch(...)->chain([...])`, `Job::withChain([...])`, and a chain on a pending dispatch held in a variable.

**A chain is ordered and a batch is not**, and the drawing says so. Chain members carry their position and arrows run through them in the order they execute; batch members carry no position, because a batch makes no promise about which runs first and drawing one would invent a fact.

**A member the graph does not hold is skipped, but its place is still counted.** Reading `1, 3, 4` says something true about a chain whose second job is filtered out of this tab; renumbering them `1, 2, 3` would say the second job runs after the first, and it does not.

**An entry nobody can read stops the group, it does not fake it.** `Bus::chain([$job, new NotifyWarehouse])` has one member that cannot be named from the source, and it is left unnamed rather than guessed.

## Two defects found by rendering it, not by testing it

**Region labels were written on top of each other.** This change is what makes a node able to sit in several regions at once, and every region anchored its name to the same corner of the node — so a job dispatched in a chain from inside a transaction had two names rendered at one point and neither was readable (right-hand screenshot). Overlap is now resolved by proximity rather than by an identical anchor: a hull label and a member label for the same node start a few pixels apart, so keying on the exact point still left both on line zero. The left screenshot has a job sitting in three chains, a batch and two transactions, all six names legible.

**Stated plainly: there is no frontend test harness in this repository**, so that fix is pinned by a render and a screenshot rather than by a test. I did not want to introduce vitest inside this PR without asking.

## What it finds on a real application, which is nothing

Measured against a 60-module application: **12 transaction regions, 0 chains, 0 batches** across 623 tabs. That is the correct answer, not a gap. The application has exactly one `Bus::batch`, and it dispatches a variable built by a collection pipeline — the job class is knowable to a reader but not from the argument list, and inferring it would mean following a `->map()` through a collection. The screenshots here are from the fixture project this branch adds, which exercises all five forms.

## Verification

`pint` clean, `phpstan` no errors, **452 tests passing**.

The rebase onto the fixed #117 had one real conflict, resolved in favour of this branch: #117 pinned the span-pairing fix on a single `data['transactionId']` field, which this change replaces with a list of regions so a node can sit in *every* span that reached it. #117's three tests are rewritten against the new shape rather than deleted — the invariant survives and gets stronger, since there is no longer a single field for one span's identity to share with another span's marking. Removing the per-span kind turns two of them red.


SanderMuller (changes requested):
Reviewed at `da9062b`, the whole diff — all ten commits, including #117's. There is no way to approve part of a branch, so the three commits you inherited are reviewed here as part of this one.

`composer test` in a fresh worktree: **452 passed (1238 assertions)**, PHPStan **No errors**, Pint clean, 15/15 CI jobs. Your numbers.

The design is right and the drawing is the good kind of hard-won. One thing stops it, and it is the finding I raised on #117 — still there, and this PR is what makes it visible.

## Blocking: the region is drawn around the method that opened the transaction, and nothing else

`CallChainEdge` is constructed at four places. Three of them are in `MethodTracer`, and only two pass `transactionId`:

- `MethodTracer.php:184` — the invokable-controller path — passes it.
- `MethodTracer.php:350` — `traceDeep` recursion — passes it.
- **`MethodTracer.php:256` — the controller-method path — passes `inTransaction` and `inRollback`, and not `transactionId`.**

The tracer computes the id correctly; it is dropped on the way out. Instrumenting the stamping loop and then the consumer on a two-controller fixture:

```
STAMP App\...\OrderController::store -> App\Services\OrderService::createOrder inTx=1 span=0 key='App\...\OrderController::store#0'
EDGE  App\...\OrderController::store -> App\Services\OrderService::createOrder inTx=1 inRb=0 txId=NULL
```

`store()` wraps `$this->service->createOrder()` in `DB::transaction(...)`. The service node comes out `inTransaction: true`, `regions: null` — flagged as inside a transaction, drawn inside no box. Both controller shapes go through this site, so it is every transaction opened in a controller method.

On the three applications I gate against, as submitted:

| | nodes flagged `inTransaction` | of those in a region | **in no region** | transaction regions | **with exactly one member** |
| A | 40 | 13 | **27** | 12 | **11 of 12** |
| B | 73 | 54 | 19 | 31 | 12 of 31 |
| C | 17 | 14 | 3 | 8 | 5 of 8 |

Eleven of A's twelve transaction regions are a box drawn around one node — the method that opened the span — with everything it wraps outside the box and flagged instead. That is the picture this PR exists to draw.

**One line fixes it**, at `MethodTracer.php:256`:

```php
inTransaction: $hop['inTransaction'] ?? false,
inRollback: $hop['inRollback'] ?? false,
transactionId: $hop['transactionId'] ?? null,   // ← add this
```

I applied exactly that and re-ran everything:

| | in no region, before → after | single-member regions, before → after | region count |
|---|---|---|---|
| A | 27 → **0** | 11 of 12 → **1 of 12** | 12 → 12 |
| B | 19 → **0** | 12 of 31 → **3 of 31** | 31 → 31 |
| C | 3 → **0** | 5 of 8 → **3 of 8** | 8 → 8 |

The region *count* does not move on any of the three. It adds members to the regions that already exist rather than inventing new ones, which is what tells you it is the right fix and not an over-correction. The whole suite stays green with it applied — **452 passed, unchanged** — which is the other half of the problem: nothing pins the omission in either direction. A test that traces a controller method wrapping a call in `DB::transaction()` and asserts the callee shares the opener's region id would have caught this on #117 and would catch it again.

That is what would change my mind: the line, and a test that fails without it.

## Three more things to fix while you are in here

**`.gitignore` no longer ignores the docs build output.** `docs/.vitepress/dist/` became `docs/.vitepress/dist/.DS_Store` — the last line had no trailing newline and `.DS_Store` landed on the end of it rather than under it. Checked both trees against the same file:

```
base (c2bf14b):  .gitignore:10:docs/.vitepress/dist/   docs/.vitepress/dist/index.html
head (da9062b):  -> index.html NOT ignored
```

So the whole VitePress build directory is now untracked noise and only a stray `.DS_Store` inside it stays ignored. Two lines instead of one:

```
docs/.vitepress/dist/
.DS_Store
```

**A chain entry nobody can read loses its place, which is the thing the description says it does not do.** The write-up draws the distinction well — a member filtered out of the tab keeps its index so the survivors read `1, 3, 4`. That holds, and `GraphBuilderRegionsTest.php:137` pins it. The other case does not: for `Bus::chain([$job, new NotifyWarehouse])` the tracer records a dense list — your own `MethodTracerJobGroupsTest.php:105` asserts `['App\Jobs\NotifyWarehouse']` — so `stampJobGroupRegions()` reads the array index and hands the survivor position 0. Measured through your own helpers:

```
PROBE regions = {"chain":{"id":"...::chainWithAnEntryNobodyCanRead#chain0","kind":"chain","position":0}}
```

The graph says `NotifyWarehouse` is the first job in that chain. It is the second, and the first is the one that could not be named. Keeping the placeholder in the list — an empty string, which `stampJobGroupRegions()` already skips — would make both cases behave the way the description says they do.

**A stale bundle chunk is committed.** The branch adds two entry chunks, `index-C7qPz7z5.js` and `index-DTbfyBoa.js`. The blade names the first; a grep across `resources/` and `frontend/` finds no reference to the second at all. It is what an intermediate rebuild left behind. `npm ci && npm run build` reproduces the tree exactly (`git status` empty), so the fix is `git rm` on the orphan, not another build. Worth catching now — this directory has reached 405 files that way before.

Also: this collides with #128 and #129, which both rewrite line 11 of `index.blade.php` and delete `index-B5C0iocj.js`. Whichever of the three merges last resolves with `npm run build` and commits what it writes, never by picking a hash out of the conflict.

## What I checked and found nothing wrong with

**The output gate**, null arm first — base against base on all three applications, identical. Then the branch: **node counts, edge counts and tab counts unchanged everywhere**, no tab added, removed or altered, and the only fields that move on any node are `data.inTransaction` and `data.regions`. For a change of this size that is a remarkable result, and it is the reason the blocking finding above is the only one in the analysis: everything else the branch touches, it touches without disturbing.

**Mutations**, whole suite each time, anchors asserted, restores verified:

| mutation | result |
|---|---|
| give a batch positions too | 1 failed |
| let the rollback path lose its own kind | 4 failed |
| keep only the last span a node was reached from | 1 failed |
| ignore the `job_groups.enabled` switch | 4 failed |

The span-pairing invariant you rewrote against regions does bite — that is the third row, and it is the one #117 got wrong.

**Renumbering chain positions densely survives the suite**, and I chased it rather than reporting it: the mutation is inert, because the list is already dense by the time `stampJobGroupRegions()` sees it. The defect is upstream in the tracer, which is the second finding above rather than a gap in your tests.

**The `job_groups` switch** is a real switch in both directions — the mutation that ignores it fails four tests, and the config path is exercised by `ProjectAnalyzerJobGroupsConfigTest`.

**On one of the three applications the branch finds a chain** — one chain, two members — alongside 31 transaction regions. Your write-up reports 0 chains and 0 batches on the application you measured, so the reading does fire on real code somewhere; it may be worth a line, since "correctly nothing" reads differently from "nothing yet".

## On the frontend test point

You said plainly that there is no frontend test harness here and did not want to add vitest inside this PR. That is the right call on vitest, and the honesty is appreciated. For what it is worth, the components can be run in this repository without adding anything to it: bundle a probe with the project's own esbuild (`--jsx=automatic`), then either `react-dom/server`'s `renderToStaticMarkup()` for a static assertion or `jsdom` plus `react-dom/client` and `act()` for a real click. I used the second of those on #128 to check a sidebar filter, and the first on #129 to check two legend rows. The label-stacking fix could be pinned that way — feed two regions whose labels anchor within a few pixels of each other and assert the resolved offsets differ — without a test runner in `package.json`.

Not asking for it in this PR. Mentioning it because "pinned by a render and a screenshot" is a smaller gap than it looks.

---

Requesting changes for the transaction-region line. Everything else in this branch — the region-as-a-kind generalisation, the five dispatch forms, the rollback path getting its own kind rather than a colour, the layout work — reads as careful and measures clean.

webard:
All four addressed, in `3980453`. Thank you for reviewing the inherited commits as part of this — that is exactly where the blocking one lived, and reviewing the branch as a whole is what made it visible.

## The blocking one — fixed on #117, and this branch now carries it

You raised it on #117 and I fixed it there before this review landed, which is why you still saw it: this branch was based on #117 at `dcfa706`, one commit before. Rebased onto `e657389`, so the line is here now, together with the test you asked for.

That test (`TransactionSpanReachesTheNodeTest`, added on #117) runs a fixture through the real pipeline and asserts the span identity reaches both the edge and the node. It is deliberately shaped to touch **all three** construction sites, so the omission is pinned in either direction:

```
no mutation              3 passed
remove at traceClosure   1 failed, 2 passed
remove at trace          2 failed, 1 passed
remove at traceDeep      1 failed, 2 passed
```

Its third assertion is restated on this branch: #117 pins the invariant on a single `transactionId` field, which this change replaces with a region list, so here it asserts the node names at least one span of kind `transaction` or `rollback`. Same invariant, new shape.

Measured on the application I gate against — smaller than yours, since it traces few transactions, but the number you care about is clean:

| | flagged `inTransaction` | in no region |
|---|---:|---:|
| before | 6 | 1 |
| after | 6 | **0** |

## `.gitignore` — fixed at the source, on #117

The fused line came from `0519baf`, this branch's own first commit on #117, so I fixed it there rather than here: `docs/.vitepress/dist/` and `.DS_Store` on separate lines. `git check-ignore` now resolves `docs/.vitepress/dist/index.html` to line 10 again.

## The chain entry nobody can read — you were right, and the description was wrong

`jobsInArray()` dropped the entry rather than holding its place, so the survivor reported the position of the one that could not be named. It now keeps an empty placeholder — which `stampJobGroupRegions()` already skipped, so that half was waiting for it. Both hop loops skip it too, so the placeholder names no node.

One thing worth reporting, because it nearly went in green: my first version of the test traced `chainWithAnEntryNobodyCanRead` *and* `chainThroughTheFacade`, and passed with the fix reverted — `NotifyWarehouse` sits at position 1 of the second chain too. Narrowed to the one method, it fails without the placeholder and passes with it. Two existing assertions were pinning the old behaviour (`JobGroupsTest` and `MethodTracerJobGroupsTest`) and now read `['', 'App\Jobs\NotifyWarehouse']`.

## The stale chunk — removed

`git rm` on `index-DTbfyBoa.js`, no rebuild. It was left by an intermediate build of mine while fixing the label collision. Confirmed the way you did: `npm run build` afterwards leaves `git status` showing only the deletion, so the committed tree is exactly what the source produces.

Gate: `pint` clean, `phpstan` no errors, **456 passing**.

SanderMuller (approved):
Re-reviewed at `3980453`, against `da9062b` — the head my last review described. The branch was rebased onto the fixed #117 and gained one commit; net patch against net patch, 13 lines left the old patch and 255 joined the new one, so the rebase is clean and the new work is `3980453` plus what came down from #117.

All three findings are fixed. I re-ran the probes that produced them rather than reading the diff.

## The blocking one

**The two-controller probe**, the one that produced the finding:

```
before   OrderService::createOrder   inTransaction=yes   regions=null
after    OrderService::createOrder   inTransaction=yes   regions=[{id: '…OrderController::store#0', kind: 'transaction'}]
```

**And on three applications**, null arm first — the previous head against itself, identical:

| | nodes flagged `inTransaction` | in no region | after | regions | holding one node |
|---|---:|---:|---:|---|---|
| A | 40 | 27 | **0** | 12 → 12 | 11 → 1 |
| B | 73 | 19 | **0** | 31 → 31 | 12 → 3 |
| C | 17 | 3 | **0** | 8 → 8 | 5 → 3 |

Those are the same numbers I got applying the one-line change by hand last time, which is the answer I wanted: the fix that shipped is the fix that was measured. The only field that moves on any node across all three is `data.regions`; no node, edge or tab is added, removed or otherwise touched, and the region count is identical everywhere, so memberships were added and no region was invented.

One node on B is worth calling out because it is the shape in miniature: a model that already sat in three transaction spans now sits in a fourth — a controller's, the one the missing id had been dropping. Nothing lost, one membership gained.

## The other two

**A chain entry nobody can read keeps its place.** Measured through your own helpers:

```
before   position: 0     ← the graph said it runs first
after    position: 1     ← it runs second, and now says so
```

Fixing it in `jobsInArray()` rather than at the stamping end is the right place — the list is what carries the order, and every consumer already knew to skip an empty entry. The note in the commit message about narrowing the test after the first version passed with the fix reverted is the part I would keep; that is the check most people skip.

**The orphan chunk is gone**, one entry chunk remains, the blade names it, and `npm ci && npm run build` leaves `git status` empty. **`.gitignore`** is fixed on #117 and comes down with the rebase; `docs/.vitepress/dist/index.html` is ignored again here.

**Mutations on the new work**, whole suite each, anchors asserted, restores verified: dropping the placeholder fails 3 tests, dropping the second hop-loop guard fails 1.

Suite green on this head: **456 passed (1248 assertions)**, PHPStan no errors, Pint clean.

## One test-strength note, non-blocking

The commit message says the placeholder "names no node" because both hop loops skip it. Only one of the two loops is watched. Removing the first guard — the one in `foreach ($group->jobs() as $position => $class)` — leaves the suite green, and I chased that rather than reporting it, because a green suite after a mutation means one of two opposite things. It is the bad one:

```
as shipped                  job | app_jobs_notifywarehouse::handle
                            service | app_services_shipmentdispatcher::chainWithAnEntryNobodyCanRead

first guard removed         job | ::handle                    ← a nameless job node on the canvas
                            job | app_jobs_notifywarehouse::handle
                            service | app_services_shipmentdispatcher::chainWithAnEntryNobodyCanRead
```

A phantom node with an empty class, and 456 tests still pass. The guard is load-bearing and nothing is watching it. One assertion in the existing `chainWithAnEntryNobodyCanRead` test — that the graph holds exactly two nodes, or that no node id starts with `::` — closes it. I mutated the two guards separately for exactly this reason; together they would have looked pinned.

## Verdict

Approving. Three findings raised, three fixed, each re-measured with the probe that produced it, and the region fix lands on the number I predicted while leaving the region count untouched on all three applications. The nameless-node gap is a test to add, not a defect to fix.

PR #133, opened 2026-09-02
Feat: show the methods an application adds to classes that do not declare them

webard:
A macro is invisible by construction. `$table->money()` resolves to nothing in `Blueprint`, and no amount of reading that class explains where the method came from — the registration is somewhere else entirely, usually a service provider, and the only way to find it today is to already know it exists.

<table>
<tr>
<td></td>
<td></td>
</tr>
<tr><td align="center"><sub>Receiver → method → the file that creates it</sub></td><td align="center"><sub>The panel, down to the line</sub></td></tr>
</table>

## The method becomes a thing with a name

Grouped by receiver, because that is the question's shape: nobody wonders "what macros exist", they wonder where `$table->money()` came from — and the receiver is the half they already have in front of them. The panel answers it outright: **adds to** `Illuminate\Database\Schema\Blueprint`, **registered by** `Core\Providers\BlueprintMacrosProvider`, and the file and line in the header.

The point of a node rather than a bullet on the provider's panel is that a node is **searchable by the method name**. Typing `money` leaves that one node at full opacity and dims the other fifty.

Read from the registration, not from a call chain, so coverage does not depend on the tracer reaching anything — the same reason the broadcasting pass is built this way.

## Detection keys on the call, never on the receiver's traits

Filament ships its **own** `Macroable` (`Filament\Support\Concerns\Macroable`), separate from Laravel's, with identical `macro()` and `mixin()` signatures. A check for `Illuminate\Support\Traits\Macroable` would silently drop every Filament component macro — **13 of the 32 direct registrations** on the application measured here.

Mixins are expanded into the public **instance** methods they contribute, in both argument spellings (`mixin(new X)` and `mixin(X::class)`). Protected helpers, statics and constructors are not methods the receiver gains, and reporting them would send a reader looking for something that does not exist.

A `macro()` whose name is a variable is a real registration whose method name is decided at runtime. It is left out rather than filed under the variable's name.

## Where the line is drawn

**Where a method is defined can be known; who calls it cannot.** Attributing `$table->money()` back to this macro means proving `$table` is a `Blueprint`, and those call sites live in migrations the scan does not read. Definitions are reported; uses are left alone.

## Its own tab, for the reason this keeps recurring

A macro hangs off a provider, and providers are not what tabs are seeded from — so the nodes would be built, correct, and in no tab anyone opens. The registrar edge points **from** the macro rather than at it, because the tab grows forward from the receiver groups; drawn the other way the class that creates the method sits outside the one tab that is about macros.

## Measured

On a 60-module application: **32 macros across 13 receivers, registered by 6 providers.** Receivers include `Blueprint` (13), Filament's `Column` (5), `Str`, `Collection`, `QueryBuilder`, and — unexpectedly — `BelongsToMany`.

One correction worth recording, because it is the kind of thing this pass exists to prevent. A grep for `::mixin(` on that application returns 19 hits, and I quoted that number before writing the analyzer. **All 19 are docblock prose and `@mixin` annotations; there is not one real mixin call.** The AST pass reported zero and was right. Mixin support is here because other applications use it, and on this one it finds nothing.

## Verification

`pint` clean, `phpstan` no errors, **417 tests passing**.

Five mutations, all killed: a mixin contributing protected and static methods, an unreadable macro name reported under the variable, the `X::class` mixin spelling dropped, the registrar edge reversed, and every receiver marked as vendor.

A config switch (`laravel-brain.macros.enabled`, default on) gates the analyzer's construction rather than its output, and `macros.paths` names the directories scanned — globs expanded, so a modular monolith points it at a wildcard segment. Note that the default `['app']` finds nothing in a modular monolith; that is the same shape as the other path-based passes.

**Stated plainly:** the tab opens at 17% zoom on that application, because 51 nodes lay out in one wide band. It is readable once you zoom, and the layer wrapping on #118 is what fixes the first impression. There is also no frontend test harness in this repository, so the panel section and the type registration are verified by rendering and screenshot rather than by test.


SanderMuller (approved):
Reviewed at `bdedac3`. `composer test` in a fresh worktree: **417 passed (1149 assertions)**, PHPStan **No errors**, Pint clean, 15/15 CI. Your number.

Screenshots are in the body, so nothing to ask for there.

## What it finds, on three applications that are not yours

The gate is purely additive — one `macros` tab per application, no node changed in place, nothing removed, no existing tab touched:

| | macros | receiver groups | of which from mixins | receivers the application owns |
|---|---:|---:|---:|---:|
| A | 3 | 2 | 2 | **0 of 2** |
| B | 56 | 16 | 54 | **0 of 16** |
| C | 22 | 9 | 19 | **0 of 9** |

Every one of the 27 receiver groups is flagged `vendor: true`, and correctly — the receivers are `Blueprint`, `Str`, `Carbon\CarbonInterval`, `Illuminate\Http\Request`, Filament components and the like. That is the case your comment calls the interesting one, and on real code it is not the interesting case, it is *every* case.

**Two things there that are worth telling you.** Your application had 32 macros across 13 receivers; B has 56 across 16, so this is not a small pass on a large codebase. And **54 of B's 56 come from mixins**, with 19 of C's 22 — where your own application had, as you found, not one real mixin call. The half you built speculatively is the half doing the work on two of my three.

Filament receivers turn up on C, so the "key on the call, not on the receiver's traits" decision pays for itself outside your codebase too.

## Cost, which is the conversation this PR needs

The benchmark comment at this head: **+57.8% (±0.4%)** on the large scan, **+58.2% (±1.1%)** on the small, `graph` phase **+273.6%**, and `parse()` calls 636 → 1,187. The generated corpus contains no `::macro(`, no `mixin(` and no `Macroable`, so that is the **idle** cost — what an application pays to be told it has no macros.

On real applications it is worse, measured by toggling your own switch in one process so nothing else varies:

| | `macros.enabled` off | on | |
|---|---:|---:|---|
| A | 1.98 s / 733 parses | 2.49 s / 1,163 | +26% |
| B | 1.46 s / 1,098 | 2.47 s / 2,040 | **+69%** |
| C | 1.54 s / 746 | 4.52 s / 4,372 | **×2.9** |

The off arm reproduces the base scan's parse counts exactly on all three, so the switch is a real switch. The cost is entirely files the rest of the scan never opened — C parses 746 files to build its graph and 4,372 to answer this question.

**I thought the default path was too broad and I was wrong, which is worth saying because it removes the obvious fix.** `macros.paths` defaults to all of `app`, and since your own application had all 32 registered by 6 providers, narrowing it to `app/Providers` looked like most of the cost for none of the loss. Measured:

```
paths = ['app']            A 3   B 56   C 22        parses  1,163 / 2,040 / 4,372
paths = ['app/Providers']  A 1   B  2   C  3        parses    733 / 1,098 /   746
```

Narrowing costs nothing at all — providers are already parsed — and loses 54 of B's 56 macros. So the broad default is right, the cost is not an accident of configuration, and there is no cheap version of this pass.

### The part that changes the arithmetic

#131's reachability pass and this one open **the same files**. I merged the two branches in a scratch worktree (the conflicts are frontend-only; the PHP merges clean) and measured four arms:

| | wall | real parses |
|---|---:|---:|
| B, neither | 1.70 s | 1,098 |
| B, macros only | 2.17 s | 2,040 |
| B, reachability only | 2.24 s | 2,110 |
| **B, both** | **2.50 s** | **2,110** |

| | wall | real parses |
|---|---:|---:|
| C, neither | 1.71 s | 746 |
| C, macros only | 4.51 s | 4,372 |
| C, reachability only | 4.49 s | 4,416 |
| **C, both** | **5.21 s** | **4,416** |

The parse count does not move at all when the second feature is added. Naive addition predicts +60% on B and roughly 7.3 s on C; the measured answers are +47% and 5.2 s. **Whichever of these two ships first pays for the parse pass, and the second costs about a quarter of what its own benchmark comment says.**

That cuts both ways and it is the maintainer's call which way. If this ships on, turning #131 on afterwards is much cheaper than its own +55% suggests. If neither ships on, both remain individually expensive to enable. What is not true is that the two costs add up, and the benchmark comments cannot show that because each measures against the same merge base alone.

### So what I would ask

Not for a change to the code. The switch exists, it is a real one, and the analysis is correct. What I would want is the C row above in the description, next to CI's +57.8% — a reader deciding whether to leave this on should see that the worst case measured is a scan taking three times as long, and that it is the modular-monolith shape that pays it. #131 ended up defaulted off on a smaller version of exactly this argument; whether this one follows is yours and the maintainer's to settle, and unlike reachability's list this output is 22 to 56 precise rows rather than 90% of everything declared, which is a real argument for keeping it on.

## Mutations

Your five, whole suite each, anchors asserted in Python with the diff printed beside the result:

| mutation | result |
|---|---|
| a mixin contributes protected and static methods too | 4 failed |
| an unreadable macro name reported under the variable | 4 failed |
| the `X::class` mixin spelling dropped | 5 failed |
| the registrar edge reversed | 2 failed |
| every receiver marked as vendor | 1 failed |

All five, as you said.

## Two small things on the viewer

**Two of the four colour maps were updated.** `ACCENT_COLORS` and `ACCENT_COLORS_LIGHT` carry `macro` and `macro_group`; `BG_COLORS` and `BG_COLORS_LIGHT` do not. Every consumer falls back (`BG_COLORS[type] ?? '#0d1117'`), so nothing breaks — but on a tab of 51 macro nodes, every card renders on the plain default background while every other node type in the application gets its tinted body. Four lines, and it is the same shape as the registry gap on #129, one map short of complete rather than six.

**The group panel shows its most interesting fact as a raw property row.** Selecting a macro gives a written section — *adds to*, *registered by*, *how: mixin of X*, which reads exactly right. Selecting the receiver group gives `fqcn`, `count`, and `vendor | true` in the generic list. Your own comment says the out-of-application receiver is the case worth drawing attention to, and it is the one rendered as a boolean.

## Checked and fine

`ALL_TYPES` in `App.tsx` carries both new kinds, so they are filterable and their edges are visible — the #129 problem does not recur. The three type lists that omit them (`ComplexityPanel`, `exportUtils`, `sequenceUtils`) are the same three that legitimately omit every non-code kind. The bundle reproduces, one entry chunk in and one out, no orphan, stylesheet untouched. The macro panel renders both spellings correctly: `macro(), one method at a time` and `mixin of <class>`.

## Verdict

Approving. The pass is correct, it finds real work on three applications that are not the one it was built against, the mutations bite, and the gate is additive. The cost is large and the switch is real; the C row belongs in the description, and the default is a decision rather than a defect.

mrmarchone:
Conflicts @webard

PR #134, opened 2026-09-02
Keep the first facade-scan read for the second pass

SanderMuller:
## What

The facade scan's second pass exists so a child that never names `Facade` is still found when it extends an app-level base that does. After reading a file in round one it discarded the bytes and opened every remaining file again the moment a chain name turned up.

Completeness does not need that second open. The first pass already has the source.

## Cost

Medians of four interleaved runs in one session, warm, against this branch's merge-base. Lifecycle runs before facades and is the canary. Parse counts do not move.

| application | php files in app/ | facades | facade phase | scan |
|---|---:|---:|---:|---:|
| A | 1,121 | 0 | 59 → 66 ms | 1,996 → 2,027 ms |
| B | 1,951 | 6 | 208 → 152 ms | 1,433 → 1,371 ms |
| C | 4,309 | 0 | 445 → 441 ms | 1,540 → 1,528 ms |

B is the application that has facades, so it is the one that actually runs the second pass. That is where the 56 ms comes from. A and C have none; their facade phase remains a single read of every PHP file under `app/`.

During the phase the unread sources occupy about 6 MB / 6 MB / 16 MB. Peak RSS of the whole scan is unchanged on A and B, and +2 MB on C; the graph phase still dominates it.

## Why this shape

#80 left this phase reading every file once to apply the filter. Parallel reads did not help. A filename filter on `*Facade.php` misses application facades that are not named that way.

The `Facades\` skip is unchanged. It no longer copies each file to do it; compared with `str_replace('Facades\\', '', $code)` it admits the same 3 / 15 / 8 files on 1,121 / 1,951 / 4,309 files under `app/`.

## Gate

Full build output — graph, all subgraphs, manifest, totals, the tracer's edge list — is **byte-identical on all three applications**, 40.2 MB, 16.4 MB and 20.4 MB of dump apiece. Parse counts 733 / 1,098 / 746 on both arms.

## Tests

`397 passed (1107 assertions)` · PHPStan level 5 + larastan **No errors** · Pint clean.

No new tests. Removing the second pass fails `it discovers a facade whose own file never mentions Facade`. Re-reading the pending files from disk leaves the suite green, so the speed claim is the wall-clock table, not an assertion.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/Analysis/FacadeAnalyzer.php
+++ b/src/Analysis/FacadeAnalyzer.php
@@ -77,15 +77,19 @@ final class FacadeAnalyzer
             false,
         );
 
-        // Round one: only files that name Facade at all. A facade reaches Illuminate's Facade
-        // through `extends`, and naming that class — imported, aliased or fully qualified —
-        // leaves the token behind.
+        // One read per file. The second pass used to throw the bytes away and open every
+        // remaining file again whenever a chain name turned up.
+        /** @var array<string, string> path => source, files not yet parsed */
         $pending = [];
         foreach ($files as $path) {
-            if ($this->mightDefineFacade($path)) {
+            $code = @file_get_contents($path);
+            if ($code === false) {
+                continue;
+            }
+            if ($this->codeMightDefineFacade($code)) {
                 $this->scanFile($path, $registry);
             } else {
-                $pending[] = $path;
+                $pending[$path] = $code;
             }
         }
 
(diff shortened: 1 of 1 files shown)


SanderMuller:
The counts are the result this job can give: nothing the scan records moved, including parse calls.

The timing is idle. `generate-corpus.php` never writes a class under `app/` that extends `Illuminate\Support\Facades\Facade`, so the second pass this change is about does not run. The facade phase stays a 15 ms first-read of the tree, inside a ±16% floor. Applications A and C in the body are that shape. The 56 ms is B, which has six facades and actually walks the pending files.

